PR #5, opened 2020-12-13
UI feature

NdibeRaymond:
This feature addresses the user interface part of the project. I will submit the screenshoots when i wake up

** Update **

Below are google drive links to the UI screenshots of what we currently have (using google drive because the images will cluter this place)

Mobile: https://drive.google.com/drive/folders/1MnbshcB6JApgo83WlStC4EkfCrGQrvRg?usp=sharing

Desktop: https://drive.google.com/drive/folders/1ndumONj6HZPTtEDkWhCQP09Ot3OGO7Hr?usp=sharing


NdibeRaymond:
Update:
 I redesigned the project details page. Here is the current look

new_mobile_details_page_ui
new_desktop_details_page_ui

tuxology:
This is great! One question. If the Video is not provided by the user, where does the right vertical bar (claps, bookmark and views) get aligned to? With the image?

tuxology (approved):

NdibeRaymond:
> This is great! One question. If the Video is not provided by the user, where does the right vertical bar (claps, bookmark and views) get aligned to? With the image?

This is an important question. I haven't implemented it. But I will make it that if a video is not provided, we pick one of the images and display in place of the video. In that case, the like and save is aligned with that image. Will make that change now

tuxology (approved):

Issue #54, opened 2021-01-06
Refactoring: Code styling and Code issues (functional components + styles separation)


NdibeRaymond:
This issue addresses the need to switch from class-based components to functional components, take a more intuitive and organized approach in structuring the frontend, moving components and views styling to different files, and other minor code refactoring issues.

**Expected Outcome After Issue is fixed**
 * All class-based components should be converted to functional components
 * Change the general folder structure of the frontend to a more intuitive approach
 * Separating styles to their own different files with easy to locate file names
 * Other minor but important refactoring issues

PR #67, opened 2021-01-08
Code Refactor

NdibeRaymond:
Issue fixed: #54

What was done:
 * All class-based components was converted to functional components
 * Change the general folder structure of the frontend to a more intuitive approach
 * Separated styles to their own different files with easy to locate file names
 * Other minor but important refactoring issues


tuxology:
@kamthamc Requesting if you could take a look and leave some comments 🙏🏽

srish:
@kamthamc Also, if you could suggest us linters/ prettiers and anything that we could do to improve code readability that would be super helpful (:

zubhub_frontend/zubhub/src/components/button/Button.js
```diff
@@ -0,0 +1,52 @@
...
+
+import classNames from "classnames";
+
+import { makeStyles } from "@material-ui/core/styles";
+import Button from "@material-ui/core/Button";
+
+import styles from "../../assets/js/styles/components/button/buttonStyles";
+
+function CustomButton(props){
+  const classes = makeStyles(styles)();
```
kamthamc: I believe it's not the way to use styles from material UI.
makeStyles should be called outside of the component to create a hook and use the hook inside otherwise hook will be created everytime CustomButton is rerendered.
https://material-ui.com/styles/basics/#hook-api

It applies to all the places it's being used like this.

zubhub_frontend/zubhub/src/views/PageWrapper.jsx
```diff
@@ -0,0 +1,288 @@
...
+
+        <div className="footer-left"></div>
+        <a href="https://unstructured.studio">
+          <img
+            src={unstructuredLogo}
+            className={classes.footerLogoStyle}
+            alt="unstructured-studio-logo"
+          />
+        </a>
+        {scrollTop(props)}
```
kamthamc: Just curious about why its moved to function (specifically inside the component itself) as I don't see much value.
```jsx
<Zoom in={useScrollTrigger}>
         <div
           onClick={handleScrollTopClick}
           role="presentation"
           className={classes.scrollTopButtonStyle}
         >
           <Fab color="secondary" size="small" aria-label="scroll back to top">
             <KeyboardArrowUpIcon />
           </Fab>
         </div>
       </Zoom>
```

zubhub_frontend/zubhub/src/views/create_project/CreateProject.jsx
```diff
@@ -0,0 +1,684 @@
...
+
+        for (let index = 0; index < project_images.length; index++) {
+          upload(project_images[index]);
+        }
+      }
+    }
+  };
+
+  const mediaFieldsValidation = () => {
+    let image_upload_button = document.querySelector("#image_upload_button");
```
kamthamc: I believe react refs are better here instead of `querySelector`

zubhub_frontend/zubhub/src/views/create_project/CreateProject.jsx
```diff
@@ -0,0 +1,684 @@
...
+        "video",
+        "you are required to submit a video url here"
+      );
+      result["video_value_not_url"] = true;
+    }
+
+    if (media_fields.files.length < 1) {
+      result["image_is_empty"] = true;
+      if (video.value === "") {
+        image_upload_button.setAttribute(
```
kamthamc: I am not sure if the styles will be overridden or not if a re-render happens in the component

zubhub_frontend/zubhub/src/views/email_confirm/EmailConfirm.jsx
```diff
@@ -0,0 +1,131 @@
...
+    username: null,
+    key: null,
+  });
+
+  React.useEffect(() => {
+    let { username, key } = getUsernameAndKey(props.location.search);
+    setState({ ...state, username, key });
+  }, []);
+
+  const getUsernameAndKey = (queryString) => {
```
kamthamc: If its possible please move all the functions out of the component which are not required inside the component. As a new function will be recreated for each render.

zubhub_frontend/zubhub/src/views/email_confirm/EmailConfirm.jsx
```diff
@@ -0,0 +1,131 @@
...
+  const classes = makeStyles(styles)();
+
+  const [state, setState] = React.useState({
+    error: null,
+    username: null,
+    key: null,
+  });
+
+  React.useEffect(() => {
+    let { username, key } = getUsernameAndKey(props.location.search);
```
kamthamc: Its OK but I believe the effect is not required as the initial state can be set like this
```
const { username, key } = getUsernameAndKey(props.location.search);
const [state, setState] = React.useState({
     error: null,
     username: username ?? null,
     key: key ?? null,
   });

**Comment by NdibeRaymond** [comment]
> @kamthamc Also, if you could suggest us linters/ prettiers and anything that we could do to improve code readability that would be super helpful (:

we already use eslint and prettier in the project @srish . Or maybe you meant suggesting more strict linting rules?

**Review by kamthamc — COMMENTED** [review]
I left few comments and I saw the data from API calls is set in the react state instead of saving in redux store. I would recommend to move the api calls out of the components and use redux store.

Prettier is good, If you are using vscode you can install prettier plugin and make it as default formatter (at least for `js`)

**Comment by kamthamc** [comment]
For eslint rules I would suggest adding
https://www.npmjs.com/package/eslint-plugin-react-hooks 
https://www.npmjs.com/package/eslint-plugin-react
https://www.npmjs.com/package/eslint-config-airbnb

and enable the rules
```
"plugin:react/recommended"
"plugin:react-hooks/recommended"
"airbnb"
```

**Comment by kamthamc** [comment]
Sorry forgot to mention, 
there are multiple places my comment applies but just mentioned once for each kind of comment like

`Use react ref instead of document selector if possible`
`Prefer const over let if the variable value will not change`

File: zubhub_frontend/zubhub/src/views/PageWrapper.jsx (hunk 2 above)
Comment (NdibeRaymond) [reply zubhub_frontend/zubhub/src/views/PageWrapper.jsx]: It's just a coding choice. Nothing significant. Removing now

File: zubhub_frontend/zubhub/src/components/button/Button.js (hunk 1 above)
Comment (NdibeRaymond) [reply zubhub_frontend/zubhub/src/components/button/Button.js]: ok. Fixing it with the next PR

File: zubhub_frontend/zubhub/src/views/create_project/CreateProject.jsx (hunk 4 above)
Comment (NdibeRaymond) [reply zubhub_frontend/zubhub/src/views/create_project/CreateProject.jsx]: they will be overridden, but that is the point. form errors are not to be persisted if it can be avoided since it changes on every validation call

**Comment by NdibeRaymond** [comment]
> Sorry forgot to mention,
> there are multiple places my comment applies but just mentioned once for each kind of comment like
> 
> `Use react ref instead of document selector if possible`
> `Prefer const over let if the variable value will not change`

Noted

File: zubhub_frontend/zubhub/src/views/email_confirm/EmailConfirm.jsx (hunk 5 above)
Comment (NdibeRaymond) [reply zubhub_frontend/zubhub/src/views/email_confirm/EmailConfirm.jsx]: ok

File: zubhub_frontend/zubhub/src/views/create_project/CreateProject.jsx (hunk 4 above)
Comment (kamthamc) [reply zubhub_frontend/zubhub/src/views/create_project/CreateProject.jsx]: Form errors should not persist but their might be a chance that the errors are lost before user completely reading the errors because of a re-render

**Comment by NdibeRaymond** [comment]
hello @kamthamc , I have implemented all your recommendations. Can you review once more?

**Comment by kamthamc** [comment]
@NdibeRaymond I believe the files are not formatted can you try

Please create a file `.prettierrc.yaml`
I am just providing my prettier config but please change as needed
```yaml
printWidth: 80
tabWidth: 2
useTabs: false
semi: true
singleQuote: true
trailingComma: all
bracketSpacing: true
jsxBracketSameLine: false
arrowParens: avoid
requirePragma: false
```
and add this script to `package.json`
```
"pretty": "prettier --config ./.prettierrc.yaml --write \"src/**/*.{js,jsx}\""
```

and run the command `npm run pretty`

NdibeRaymond:
> @NdibeRaymond I believe the files are not formatted can you try
> 
> Please create a file `.prettierrc.yaml`
> I am just providing my prettier config but please change as needed
> 
> ```yaml
> printWidth: 80
> tabWidth: 2
> useTabs: false
> semi: true
> singleQuote: true
> trailingComma: all
> bracketSpacing: true
> jsxBracketSameLine: false
> arrowParens: avoid
> requirePragma: false
> ```
> 
> and add this script to `package.json`
> 
> ```
> "pretty": "prettier --config ./.prettierrc.yaml --write \"src/**/*.{js,jsx}\""
> ```
> 
> and run the command `npm run pretty`

Done @kamthamc . Review when free

tuxology (approved):

Issue #55, opened 2021-01-06
Internationalization


NdibeRaymond:
The project in its current state is not set up for localization. There is a need to internationalize the project so as to enable easy localization when the time comes

PR #74, opened 2021-01-13
Internationalization 

NdibeRaymond:
This PR concerns internationalization of the app for easy localization to multiple languages (see #55 )
What was done:
 * Internationalization of the frontend
 * Internationalization of the backend

**Image**

translate_to_hindi


zubhub_frontend/zubhub/public/locales/en/translation.json
```diff
@@ -0,0 +1,383 @@
...
+  "pageWrapper":{
+    "navbar":{
+      "login":"Login",
+      "signup":"Sign Up",
+      "createProject":"Create Project",
+      "savedProjects":"Saved Projects",
+      "logout":"Logout"
+    }
+  },
+
```
srish, line 17: May we don't need to leave two lines here and elsewhere.

zubhub_frontend/zubhub/public/locales/en/translation.json
```diff
@@ -0,0 +1,383 @@
...
+
+
+  "signup":{
+      "welcome":{
```
srish: May be this could be welcomeMsg

zubhub_frontend/zubhub/public/locales/en/translation.json
```diff
@@ -0,0 +1,383 @@
...
+      "signup":"Sign Up",
+      "createProject":"Create Project",
+      "savedProjects":"Saved Projects",
+      "logout":"Logout"
+    }
+  },
+
+
+
+  "signup":{
```
srish: May be this could be signupPage and same change can be made elsewhere in the file to other pages

zubhub_frontend/zubhub/public/locales/en/translation.json
```diff
@@ -0,0 +1,383 @@
...
+          }
+        },
+        "location":{
+          "label":"Location",
+          "errors":{
+            "min":"your location is too short",
+            "required":"this field is required"
+          }
+        },
+        "password1":{
```
srish: may be just password

zubhub_frontend/zubhub/public/locales/en/translation.json
```diff
@@ -0,0 +1,383 @@
...
+          "label":"Password",
+          "errors":{
+            "min":"your password is too short",
+            "required":"this field is required"
+          }
+        },
+        "password2":{
```
srish: this could be confirmPassword

zubhub_frontend/zubhub/public/locales/en/translation.json
```diff
@@ -0,0 +1,383 @@
...
+        "username":{
+          "label":"Username",
+          "errors":{
+            "required":"this field is required"
+          }
+        },
+        "email":{
+          "label":"Email",
+          "errors":{
+            "invalidEmail":"invalid email",
```
srish: this could just be invalid

zubhub_frontend/zubhub/public/locales/en/translation.json
```diff
@@ -0,0 +1,383 @@
...
+       },
+       "password":{
+         "label":"Password",
+         "errors":{
+           "min":"your password is too short",
+           "required":"this field is required"
+         }
+       }
+     },
+     "submit":"Login",
```
srish: Maybe submit here and elsewhere can be shown under "inputs"

zubhub_frontend/zubhub/public/locales/en/translation.json
```diff
@@ -0,0 +1,383 @@
...
+     "others":{
```
srish: If possible try to move the items in "others" where they most seem to fit. For example, signup and forgotPassword can be moved under inputs

srish: May be errors/ warnings can have a parent key

zubhub_frontend/zubhub/public/locales/en/translation.json
```diff
@@ -0,0 +1,383 @@
...
+        },
+        "inputs":{
+          "newPassword1":{
+           "label":"New Password",
+           "errors":{
+             "min":"your password is too short",
+             "required":"this field is required"
+           }
+          },
+          "newPassword2":{
```
srish: same here.. may be change this to confirmNewPassword

zubhub_frontend/zubhub/public/locales/en/translation.json
```diff
@@ -0,0 +1,383 @@
...
+        "others":{
+          "errors":{
+            "unexpected":"An error occured while fetching Projects, please try again later"
+          }
+        }
+      },
+
+
+      "project":{
+        "years":"years",
```
srish: May be this is more general stuff and can be moved out so can be used in other code files as well. And, not just the one specific to project page.

srish (commented):
@NdibeRaymond I have left a few comments. One small thing also to do: maybe leave a space after ":" everywhere.

I wonder if you thought about the translations like in here https://github.com/wikimedia/wikimedia-toolhub/blob/main/vue/src/assets/locales/i18n/en.json as opposed to nested translations?

zubhub_frontend/zubhub/public/locales/en/translation.json
NdibeRaymond: The lines are just a means of making things clearer. it can be removed but it will be more inconveniencing to spot where each page translations start and where it ends

zubhub_frontend/zubhub/public/locales/en/translation.json
NdibeRaymond: ok

zubhub_frontend/zubhub/public/locales/en/translation.json
NdibeRaymond: That will make the page's root translation name different from the name of the page itself. I used this name so the name are identical to the name of the pages they represent. Adding page is just kinda like a redundancy since we already know each translation section represents a page.

zubhub_frontend/zubhub/public/locales/en/translation.json
NdibeRaymond: This corresponds to the naming of the password fields in the forms. Some pages (like signup) demands password and confirm password fields, our auth backend demands that these be named password1 and password2. The translation names need to correspond to these fields

zubhub_frontend/zubhub/public/locales/en/translation.json
NdibeRaymond: ok

zubhub_frontend/zubhub/public/locales/en/translation.json
NdibeRaymond: submits are buttons. But I can put them under input if you want

zubhub_frontend/zubhub/public/locales/en/translation.json
NdibeRaymond: I think these are not really part of the page, hence why I placed them under others. the errors under others as well are mostly backend and general errors, so they don't belong on a particular field. I don't understand what you meant by errors should have a parent key

zubhub_frontend/zubhub/public/locales/en/translation.json
NdibeRaymond: ok

zubhub_frontend/zubhub/public/locales/en/translation.json
NdibeRaymond: I can still change it though, but I believe it makes things easier for future developers if the translation keys are named after the elements (or input fields they translate). And the name doesn't affect the translation either. Having password1 and password2 in the forms (these names are what our auth backend package requires so can't be changed), while having password and confirmPassword in the translation file doesn't make sense if you ask me

zubhub_frontend/zubhub/public/locales/en/translation.json
NdibeRaymond: Again, this can be named signupPage. I just think it makes for better dev experience if the names of elements and pages being translated matches the translation keys

NdibeRaymond:
> @NdibeRaymond I have left a few comments. One small thing also to do: maybe leave a space after ":" everywhere.
> 
> I wonder if you thought about the translations like in here https://github.com/wikimedia/wikimedia-toolhub/blob/main/vue/src/assets/locales/i18n/en.json as opposed to nested translations?

we can add space between ":" but this won't be according to our prettier rules. For the nested translations, that is kind of important because the translation needs to have some semantics. "signup.welcome.primary" makes more semantic sense than "signup.primary". It also helps in sectioning and more flexibility for future translations. eg if we need another "primary", we won't have to rename primary to primary1. What I am trying to say is that just throwing every key into the translation doesn't seem like the right way to go about it. Even the creators of the package we are using advice that we group the translations like this. This is my opinion. it can be changed, but it will make the developer experience worse for future maintainers

zubhub_frontend/zubhub/public/locales/en/translation.json
srish: Okay, i would say let's remove them

zubhub_frontend/zubhub/public/locales/en/translation.json
srish: Okay fair enough, we can drop the idea then.

zubhub_frontend/zubhub/public/locales/en/translation.json
srish: Okay, that also makes sense. Let's not do it then.

zubhub_frontend/zubhub/public/locales/en/translation.json
srish: Yes, if you also think it makes sense put them under input

zubhub_frontend/zubhub/public/locales/en/translation.json
srish: Hmm, so if they don't belong to a particular field, can we just move them of the "others" category?

PR #100, opened 2021-02-04
Improved project creation form

### LINKED ISSUE #71: Improve the "materials" field ui
It is not obvious from the project creation form UI how to add materials. I see now that we have a placeholder that says "add a material and click the + button". But, we could make more UI improvements to make it self-explanatory how one can add materials.

@tuxology can provide more input here on how the UI could be.

NdibeRaymond:
# Issue fixed: #71 

screencapture-localhost-3000-projects-create-2021-02-05-02_27_25


tuxology (approved):
Thanks! All looks well for now. We may need to review the translations alter, but this is fine for now!

PR #286, opened 2022-02-11
Notification Model and Endpoints

### LINKED ISSUE #284: Add notifications model and add/update/delete endpoints
## Description

In order for the user to see a list of the notifications they haven't seen we need to store them in a table. 

## Learning objectives

- Creating models in Django
- Creating/testing endpoints

## Todo

- [x] Create notifications model 
- Model should contain notification message, notification recipient, whether it has been viewed or not, Date added
  
- [x] Notification endpoints - Update notification, Delete notification, Get notification by user

## Useful links

- https://www.caktusgroup.com/blog/2019/02/01/creating-api-endpoint-django-rest-framework/

zorazrr:
## Notification Model and Endpoints

Create notification model and corresponding endpoints

Closes #284 

## Changes

Create notification model
- id 
- message 
- recipient 
- source
- viewed
- date

Create endpoints for 
- Get a list of all notifications
- Get a list of notifications by user 
- Update a notification by id
- Delete a notification by id

Add notification preference enum in creator model

## Requests / Responses

**Request**
GET `/api/notifications/` Returns a list of notifications
GET `/api/notifications/[user-id]/notifications/` Returns a list of notifications associated with a certain user
POST `/[notification-id]/update/` Updates a notification
DELETE `/[notificaiton-id]/delete/` Deletes a notification


NdibeRaymond:
see https://github.com/unstructuredstudio/zubhub/pull/287#issuecomment-1038313531

zubhub_backend/zubhub/notifications/models.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class Notification(models.Model):
+    id = models.UUIDField(
+        primary_key=True, default=uuid.uuid4, editable=False, unique=True)
+    message = models.TextField(blank=False, null=True)
+    recipient = models.ForeignKey(
+        Creator, on_delete=models.CASCADE, null=True, related_name="notification_recipient", blank=True)
+    source = models.ForeignKey(
+        Creator, on_delete=models.CASCADE, null=True, related_name="notification_source", blank=True)
```
NdibeRaymond, line 15: I think `source` pointing to the `Creator` model is a little bit undesirable. We need to think about the bigger picture. Notification is the place where all the information the Zubhub APP wants to communicate to the user lives. It can be a registration welcome message, email verification reminder/message (or phone verification reminder) or a whole bunch of things that are not related to another user. 
I think source should be removed entirely. All notifications are messages from Zubhub app. Any information about the reason for the notification will be included in the notification message itself. We can parse this message if we need to do fancy stuffs (like link to the profile of the user that liked your project, etc)

zubhub_backend/zubhub/creators/models.py
```diff
@@ -66,6 +76,9 @@ class Creator(AbstractUser):
     role = models.PositiveSmallIntegerField(
         choices=ROLE_CHOICES, blank=True, null=True, default=CREATOR)
     search_vector = SearchVectorField(null=True)
+    contact = models.PositiveSmallIntegerField(
```
NdibeRaymond: contact should be inside the Settings model since it's something that is not directly related to user but is more like a settings stuff.

zubhub_backend/zubhub/notifications/urls.py
```diff
@@ -0,0 +1,14 @@
+from django.urls import path
+from .views import *
+
+app_name = "notifications"
+
+urlpatterns = [
+    path('', NoticationListAPIView.as_view(), name='list_notifications'),
```
NdibeRaymond: What will this endpoint be used for? I don't see us ever needing to fetch all notifications in the database.

zubhub_backend/zubhub/notifications/urls.py
```diff
@@ -0,0 +1,14 @@
...
+    path('<uuid:pk>/update/', UpdateNotificationAPIView.as_view(),
+         name='update_notification'),
+    path('<uuid:pk>/delete/', DeleteNotificationAPIView.as_view(),
+         name='delete_notification'),
+    path('<uuid:pk>/notifications/',
```
NdibeRaymond: notification is a personal stuff and a user should only be able to fetch their own notifications. for a user to be able to fetch their notifications, they need to be logged in. If a user is logged in, we know who the user is and we can get their username and id from `requests.user`. uuid:pk is redundant here since we already have the user's information in the request since they are logged in.

zubhub_backend/zubhub/notifications/views.py
```diff
@@ -0,0 +1,56 @@
...
+from django.utils.translation import ugettext_lazy as _
+
+from rest_framework.generics import (
+    UpdateAPIView, ListAPIView, DestroyAPIView)
+
+
+class NoticationListAPIView(ListAPIView):
+    queryset = Notification.objects.all()
+    serializer_class = NotificationSerializer
+    permission_classes = [AllowAny]
```
NdibeRaymond: All of NotificationListAPIView should be removed.

zubhub_backend/zubhub/notifications/views.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class UpdateNotificationAPIView(UpdateAPIView):
+    queryset = Notification.objects.all()
+    serializer_class = NotificationSerializer
+
+    def perform_update(self, serializer):
```
NdibeRaymond, line 22: remove all of perform_update. we shouldn't be doing much here except just updating "viewed" field. and serializer can handle that well. no need to overide `perform_update`

zubhub_backend/zubhub/notifications/views.py
```diff
@@ -0,0 +1,56 @@
...
+    UpdateAPIView, ListAPIView, DestroyAPIView)
+
+
+class NoticationListAPIView(ListAPIView):
+    queryset = Notification.objects.all()
+    serializer_class = NotificationSerializer
+    permission_classes = [AllowAny]
+
+
+class UpdateNotificationAPIView(UpdateAPIView):
```
NdibeRaymond: Rename to `MarkNotificationAsViewedAPIView` or something similar

zubhub_backend/zubhub/notifications/views.py
```diff
@@ -0,0 +1,56 @@
...
+        if notification_changed(old, new):
+            result = self.request.user.save()
+            return result
+
+
+class DeleteNotificationAPIView(DestroyAPIView):
+    queryset = Notification.objects.all()
+    serializer_class = NotificationSerializer
+
+    def delete(self, request, *args, **kwargs):
```
NdibeRaymond: remove all of `delete` method too. Django can handle single actions like deletion without much overriding.

zubhub_backend/zubhub/notifications/views.py
```diff
@@ -0,0 +1,56 @@
...
+        notification = self.get_object()
+        if notification:
+            result = self.destroy(request, *args, **kwargs)
+            request.user.save()
+            return result
+
+
+class UserNotificationsAPIView(ListAPIView):
```
NdibeRaymond, line 33:
```
class CreatorNotificationsAPIView(ListAPIView):
    serializer_class = NotificationSerializer
    permission_classes = [IsAuthenticated, isOwner]

    def get_queryset(self):
        self.request.user.notifications.all().order_by("date")
```

NdibeRaymond: Also, this should be paginated, perhaps like 10 - 20 notifications per page.

zubhub_backend/zubhub/notifications/utils.py
```diff
@@ -0,0 +1,11 @@
+def notification_changed(obj, instance):
```
NdibeRaymond: maybe delete this function? it doesn't serve any purpose now. see other reviews to know why.

zubhub_backend/zubhub/notifications/serializers.py
```diff
@@ -0,0 +1,11 @@
...
+
+from .models import Notification
+
+
+class NotificationSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Notification
+        fields = (
+            'id', 'message', 'recipient', 'source', 'viewed', 'date'
+        )
```
NdibeRaymond: ` read_only_fields = ["id", "message", "recipient", "date"]`
once a notification has been created, it shouldn't be that easy to change those fields assigned to read_only_fields.
It should only be possible to change the viewed field in notifications from False to True (not from True to False).
As for source field, see other reviews as to why I removed it.

NdibeRaymond: you should overide `update` method here to check if a notification is viewed or not and if not viewed, set as viewed.

zubhub_backend/zubhub/notifications/views.py
NdibeRaymond: Also remember to add throttling (see other views for example)

zubhub_backend/zubhub/notifications/views.py
NdibeRaymond: add throttling

zubhub_backend/zubhub/notifications/views.py
NdibeRaymond: add throttling

NdibeRaymond (changes requested):

NdibeRaymond:
It feels like the whole notification stuff should be added into the creator app directly because it's too lean and depends a lot on the Creator model to be a standalone app on it's own. the Notification model can be added into creators/models.py, same with serializer and views. We shouldn't have a standalone notification app.

zubhub_backend/zubhub/notifications/models.py
ghost: I understand what you're saying but we were more concerned about getting the profile pic and username for certain notification messages. For example, if a notification stated "<user_name> started following you" we could reference that user in ```source```  and have "started following you" in the ```message```.  This would allow us to have the profile picture and username in the notification message and also link to their profile when clicked on. Do you know how we might be able to get something like the user's profile picture without using a ```source``` field?

NdibeRaymond: Simply adding the username of the user that followed another user in the notification message is enough to get all the extra interactions we need. The profile picture is auto-generated based on the username, so having the username is enough to get the profile picture. as for linking to the profile, there is a function in the frontend that parses strings and can detect the presence of sub-strings with a particular format (@username). With this we can format the incoming notification string and replace @username with `<a href="link to profile">username</a>`.
Of course if a user changes their username, then the profile can no longer be found. But this is not much of a problem since a username is not something that changes often and if it does, we can simply report that the user profile is not found. The bigger issue is restricting the notification model to just user-to-user notifications.

NdibeRaymond: I will also think more about this to see if there is a better way. But I'm sure source pointing to Creator model is not a good design.

ghost: At some point in the future, it seems logical to allow users to upload their own profile pics and in that case, this wouldn't work. Some of the concerns you've outlined with your design are also valid so we should try to find a better design that minimizes bugs and allows for feature additions in the future. 

Would it be ok to make the ```source``` field NULL for non-user-generated notifications and also add a ```notification_type``` field to the model. This would be an enum that could hold values such as ```follow```, ```like```, ```email_verification``` so that the frontend would know how to appropriately display the notification. Depending on whether the ```notification_type``` is ```email_verification``` or something different, the frontend could choose to show an icon that appropriately reflects it. There don't seem to be too many notification use cases so this enum wouldn't end up being too long.

zubhub_backend/zubhub/APIS/urls.py
```diff
@@ -8,9 +8,16 @@
     path('rest-auth/registration/', include('rest_auth.registration.urls')),
     path('creators/', include('creators.urls', namespace="creators")),
     path('projects/', include('projects.urls', namespace="projects")),
+<<<<<<< HEAD
+    path('notifications/', include('notifications.urls', namespace="notifications")),
+    path('upload_file/', UploadFileAPIView, name="upload_file"),
+    path('delete_file/', DeleteFileAPIView, name="delete_file"),
+    path('upload_file_to_local/', UploadFileToLocalAPIView,
+=======
```
NdibeRaymond: hello @zorazrr , you should fix this merge conflict when you can. for the purpose of having a consistent style for creating url endpoints, we changed all use of underscores to "-" e.g. upload_file was changed to upload-file.

zubhub_backend/zubhub/creators/models.py
```diff
@@ -82,9 +82,22 @@ def save(self, *args, **kwargs):
...
+    CONTACT_CHOICES = (
+        (WHATSAPP, 'WHATSAPP'),
+        (EMAIL, 'EMAIL'),
+        (SMS, 'SMS')
+    )
+
     creator = models.OneToOneField(
         Creator, on_delete=models.CASCADE, primary_key=True)
     subscribe = models.BooleanField(blank=True, default=False)
+    contact = models.PositiveSmallIntegerField(
```
NdibeRaymond, line 98: Thanks for fixing this @zorazrr, this is definitely better. Another suggestion, right now we send all notifications to both email and phone. With this, we can have a logic every time we send notifications that detect the user's preferred notification channels and only notify then through that channel (you can also create a task for this)

zubhub_backend/zubhub/notifications/admin.py
```diff
@@ -0,0 +1,6 @@
+from django.contrib import admin
+
+# Register your models here.
+from .models import Notification
+
+admin.site.register(Notification)
```
NdibeRaymond, line 6: I think we can improve this a bit. Right now in the admin site this will probably just be displaying the default model str method return value. That a difficult ux for admins to handle since they'll probably have no idea about the context of the notification (to who, why, from where, etc). That improvement can be made later.

zubhub_backend/zubhub/notifications/urls.py
```diff
@@ -0,0 +1,13 @@
+from django.urls import path
+from .views import *
+
+app_name = "notifications"
+
+urlpatterns = [
+    path('<uuid:pk>/update/', MarkNotificationAsViewedAPIView.as_view(),
+         name='update_notification'),
```
NdibeRaymond: when you can @zorazrr spend sometime converting all your underscore based urls to  use "-" instead. I admit it was my fault for creating a mixed pattern before. I fixed the mixed pattern by refactoring all urls to not use underscores anymore. You can help me keep to this pattern by changing these too.

zubhub_backend/zubhub/creators/models.py
ghost: In PR #306, @zorazrr is working on creating a helper function that will call one of the task functions (send_text, send_whatsapp, send_email) based on a user's channel preferences so that other functions just have to call this helper function and not worry about that detail. This function will also add the notification to the notifications model appropriately. My plan was to add this function in a utils.py file in the notifications folder. Do you feel that would be the right place for it?

ghost (approved):

PR #306, opened 2022-02-24
Notification Helper Function

### LINKED ISSUE #304: Helper Function to Choose Communication Channel
## Description

Add a helper function that can choose which communication channel should be used for the notification.

Context: Because there are 3 different communication channels (SMS/WhatsApp/Email), we have to choose which one to send on based on the user's preferences. This function will abstract that logic away so that other functions can just call this function and not worry about that preference. 

## Todo

- [x] Add helper function
  - Should accept the creator(s) to send to, context, and template name
  - Should use Settings model to choose communication channel

## Useful References

- Check detect_mentions in projects/utils.py for how notifs are currently sent

zorazrr:
## Summary

Adds a helper function `send_notification(users, contexts, template_name)` to choose notification method.

Closes #304 

## Changes
- Add helper function in `/creators/utils.py`
- Goes through an array of users and call the corresponding "send notification" function according to the user's communication preference


zubhub_backend/zubhub/creators/utils.py
```diff
@@ -237,6 +237,42 @@ def activity_notification(activities, **kwargs):
...
+                template_name=template_name,
+                ctx=context)
+
+        if user_setting.contact == Setting.EMAIL:
+            context.update({"email": user.email})
+            email_contexts.append(context)
+
+        if user_setting.contact == Setting.SMS:
+            context.update({"phone": user.phone})
+            sms_contexts.append(context)
```
ghost: Before adding their phone number or email, verify that they have one. ```detect_mentions``` has an example of this.

ghost (approved):
Great work Zora! There's just one small thing you need to fix but after that you're good to merge.

Issue #299, opened 2022-02-22
Notification Item


ghost:
## Description

Create a component for each Notification in the Notification Center

Context: This is where users can see the information regarding their notification

## Todo

- [x] Create a component that can accept things like message, profile pic, whether it was viewed so that it can be called from a map function
- [x] Redirect based on link when clicked on
- [x] Adapt for mobile

## Useful links

- Mobile Figma Design: https://www.figma.com/file/rdoLGoH5vfFBfqV11lgOyw/ZubHub-Mobile?node-id=0%3A1
- Web Figma: https://www.figma.com/file/jRQdwIwXdgmXEe5L0XLgno/ZubHub-Web?node-id=0%3A1

yaya-usman:
@ajain1921 can i be assigned this as well?

PR #311, opened 2022-02-28
Notification Component for Notification Panel

anthonycruzmacedo:
## Summary

Notification component for notification panel that displays a source's profile picture, a notification message, a timestamp, 
a conditional view dot, and redirects when clicked
Closes: #299 

## Changes

* create ```Notification.jsx```, ```NotificationStyles.js```
* add ```viewNotification``` method to ```api.js```
* add ```useParams``` to ```ProjectDetails.jsx```, update ```useEffect()``` for ```getProject``` method to fix page reload bug when a notification is clicked while already on a project page

## Screenshots
image


zubhub_frontend/zubhub/src/components/notification/Notification.jsx
```diff
@@ -0,0 +1,140 @@
...
+        </>
+      );
+    } else if (notification.type == NOTIFICATION_TYPE.FOLLOWING_PROJECT) {
+      return (
+        <>
+          <strong>{notification.source.username}</strong> posted a new project!
+        </>
+      );
+    }
+  };
```
ghost: To clean up the component a bit, can you move this to a separate scripts file like some other components do? Also ```getNotification``` has a lot of repeated logic so you might be able to simplify that by maybe storing the strings and using the ```notification_type``` to choose it.

zubhub_frontend/zubhub/src/components/notification/Notification.jsx
```diff
@@ -0,0 +1,140 @@
...
+          alt="user-profile"
+        />
+
+        <div className={classes.text}>
+          <p className={classes.message}>{getNotification()}</p>
+          <p className={classes.time}>{renderTimeAgo()}</p>
+        </div>
+        {!notification.viewed && <div className={classes.viewDot}></div>}
+        {notification.viewed && <div className={classes.unviewed}></div>}
+      </div>
```
ghost: Great work on the css! However, we want to try to use as many Material UI components as possible so that we don't have to do a lot of the styling ourselves. I quickly wrote up the following block of code

```
<ListItem
        alignItems="flex-start"
        disableGutters
        button
        className={classes.notificationStyle}
      >
        <ListItemAvatar>
          <Avatar
            src={notification.source.avatar}
          />
        </ListItemAvatar>
        <ListItemText
          primary={getNotification()}
          secondary={renderTimeAgo()}
          style={{ color: 'black', width: '275px' }}
        />
        /* notification icon div */
</ListItem>
```

based on this doc (https://v4.mui.com/components/lists/#folder-list) and you'll notice that it accomplishes a core part of the UI with very minimal styling of our own. If you use this example and some of your previous styling you should be able to achieve the same styling in a cleaner fashion.

zubhub_frontend/zubhub/src/views/PageWrapper.jsx
```diff
@@ -104,6 +105,49 @@ function PageWrapper(props) {
...
+      followers: [],
+      following_count: 0,
+      projects_count: 0,
+      members_count: null,
+      role: 'creator',
+    },
+    link: 'youtube.com',
+    viewed: false,
+    time: '2022-03-12T21:30:57.732Z',
+    // time: '2022-03-12T01:53:51.431Z',
```
ghost: I'll try to merge Andrew's PR by tonight so you can use the updated model to connect your component with the backend but nice work using mock data.

zubhub_frontend/zubhub/src/store/actions/userActions.js
```diff
@@ -357,6 +357,9 @@ export const addComment = args => {
   };
 };
 
+export const viewNotification = props => {
+  return API.viewNotification(props)
+};
```
ghost: Since this doesn't relate to the Redux state , we can simply call ```API.viewNotification()``` from the component itself and not include it here.

ghost (commented):
Great work overall! I left a few comments about how you can make the code a little cleaner. In terms of next steps, I'll get Andrew's stuff merged in so you can connect with the backend. Also, it might make sense to move ```Notification``` into ```notification_panel``` but up to you.

NdibeRaymond:
Great job overall, thanks to everyone involved! Though still draft, something caught my attention. It seems we are not building with internalization in mind. The `renderTimeAgo` function easily comes to mind as a violator. If we can we should use `dFormatter` for everything concerning date formatting and `nFormatter` for anything concerning number formatting. It is easy to hook up the values returned from these functions into i18next for translation. To get an idea, try searching for `dFormatter` and observe how it's being used across the codebase.

anthonycruzmacedo:
Hi Raymond thanks for the feedback. I've updated the component to use dFormatter for the time stamp, as well as i18next translations for the message and timestamp.

zubhub_frontend/zubhub/src/assets/js/styles/components/notification/NotificationStyles.js
```diff
@@ -0,0 +1,91 @@
+const styles = theme => ({
+    notificationStyle: {
+        display: 'flex',
+        borderRadius: '5px',
+        width: '340px',
```
ghost, line 5: On the mobile view, the notifications don't take up the entire width of the screen because the width is pre-defined here. You can use [breakpoints](https://mui.com/customization/breakpoints/) to determine when you should set the width.

zubhub_frontend/zubhub/src/components/notification/Notification.jsx
```diff
@@ -0,0 +1,99 @@
...
+          <ListItemAvatar>
+            {Object.keys(notification.sources).length === 1 && (
+              <AvatarGroup className={classes.group}>
+                <Avatar
+                  className={classes.image}
+                  src={notification.sources[0].avatar}
+                />
+              </AvatarGroup>
+            )}
+            {Object.keys(notification.sources).length > 1 && (
```
ghost: Since ```notification.sources``` is already an array, you can just get the length property from the array itself.

zubhub_frontend/zubhub/src/components/notification/Notification.jsx
```diff
@@ -0,0 +1,99 @@
...
+              </AvatarGroup>
+            )}
+          </ListItemAvatar>
+          <ListItemText
+            className={classes.text}
+            classes={{ primary: classes.message, secondary: classes.time }}
+            primary={
+              <span
+                dangerouslySetInnerHTML={{
+                  __html: t(getMessage(notification.message)),
```
ghost: The ```t``` parameter can't exactly translate any text you give it. It accepts an object reference to the objects defined in the ```public/locales``` jsons to determine what the translation is. For static text, we can use the ```t``` parameter but for dynamic text that comes from the backend we can't exactly do that yet. For now, we don't need to use it for any dynamic notification text.

zubhub_frontend/zubhub/src/views/PageWrapper.jsx
```diff
@@ -104,6 +105,49 @@ function PageWrapper(props) {
   const { t } = props;
   const { hero } = props.projects;
   const profileMenuOpen = Boolean(anchor_el);
+
+  var notif = {
```
ghost: we can remove this

zubhub_frontend/zubhub/src/components/notification/Notification.jsx
```diff
@@ -0,0 +1,99 @@
...
+    let message = '';
+    // 55 character message limit + 17 for strong tags being counted as well
+    if (notification.message.length < 72) {
+      message += notification.message;
+    } else {
+      message += notification.message.slice(0, 72) + '...';
+    }
+    return message;
+  };
+
```
ghost: [This](https://stackoverflow.com/questions/15909489/text-overflow-ellipsis-on-two-lines) might be a nice option for text overflow.

ghost (approved):
Great work Anthony! Thanks for fixing some of the templates and getting the diagonal pics to work. Overall the component looks really clean and I just left a few comments about minor fixes. Feel free to squash merge when you're done.

zubhub_frontend/zubhub/src/components/notification/Notification.jsx
anthonycruzmacedo: I looked at this a bit but it doesn't seem -webkit-box is compatible with JS styles

PR #320, opened 2022-03-07
Auto save on project forms

zorazrr:
## Summary
Closing #322 

## Changes
Autosave in "Create Project" form. Project autosaved is automatically saved into the database as a `draft`.
The core is the `timeoutSave` function. Some key things changed include:
- Plugged in `timeoutSave` in handle-change functions
- Instantly update project fields on autosave (because react setState lagging)
- Disabled validation for autosaved projects
- Dome fields were made optional in the model
- Made some validation on the backend for non-draft situations
- Disabled redirect on autosave
- Added some conditionals to switch from `createProject` to `updateProject` after the first autosave

## Note to Self: To-Dos
- [x] conditionally trigger autosave (only on create project form)
- [x] resolve unable to open projects without `title` issues
- [x] conditionally execute `validate_video` and `validate_images` in `serializers.py`
- [x] prevent autosave to create project multiple times because of undefined `props.match.params.id`
- [x] turn on backend validation for `material_used`
- [x] fix not autosaving after image upload issue


tuxology:
@ajain1921 @zorazrr Are we close to merging this? I remember that this is also almost complete

Issue #321, opened 2022-03-07
Add create notification function


ghost:
## Description

Add function to create a notification in the Notifications model.

Context: This function will be called inside Zora's function (#306) to add to the model.

## Learning objectives

- Django REST Framework

## Todo

- [ ] Add create notification helper function

PR #324, opened 2022-03-08
Add several functions and maps for notifications logic

AndrewLester:
## Summary

Notifications are now working through tested channels (DB, Whatsapp). We need to setup sendgrid + twilio SMS for further testing.
EDIT: Now tested on sendgrid and twilio and confirmed to be working.
Closes: #321 

*Questions*
- Celery kept crashing and wanted me to add the `django_extensions` dependency, which I did and it fixed the problem. That's why the requirements.txt in celery has it. Is this incorrect?
- One of the migrations for creator settings was duplicated, so I deleted it and all is well.

## Changes

- Changes `send_notification` function to call `push_notification` and properly check and send to necessary channels.
- Use the `send_notification` function in proper places in the backend.


zubhub_backend/zubhub/notifications/models.py
```diff
@@ -21,6 +22,7 @@ class Type(models.IntegerChoices):
         Creator, on_delete=models.CASCADE, null=True, related_name="notification_recipient", blank=True)
     source = models.ForeignKey(
         Creator, on_delete=models.CASCADE, null=True, related_name="notification_source", blank=True)
+    message = models.CharField(max_length=255, blank=True, null=True)
     link = models.URLField(max_length=1000)
```
ghost: When I try entering a URL through Django's Admin View like ```/projects/``` it rejects it because it has a built-in URL validator. This isn't an issue when using the endpoint but it would make more sense to use a ```CharField``` because we aren't storing proper URLs anyway. 

See this for reference: https://docs.djangoproject.com/en/dev/ref/models/fields/#urlfield

ghost (approved):
Great work on the PR! I left a comment about a small fix but everything else looks good. Before you merge can you test the email and SMS notifications?

PR #327, opened 2022-03-12
Add token with redux

ghost:
## Summary

Uses the redux state to access the authentication token. Makes API calls a little easier.

Note: It would be tedious to refactor the current API calls to reflect this change so I'm leaving them for now but API calls we write from now on can be cleaner.

## Changes

- Get token from redux state


NdibeRaymond:
This is great! this is something I've been putting off for sometime now. I am going to extend this by refactoring all the methods so we don't pass tokens through the actions anymore. I think leaving this in a half-baked state is not so great. Once I upload a patch for the refactor, then we an merge this. Thanks @ajain1921 for this

PR #349, opened 2022-03-28
Video compression

gracewzhang:
## Changes

- Item 1
- Item 2
- Item 3

## Screenshots

### DIFF
--- a/zubhub_frontend/zubhub/package-lock.json
+++ b/zubhub_frontend/zubhub/package-lock.json
@@ -13148,6 +13148,11 @@
       "resolved": "https://registry.npmjs.org/react-fast-compare/-/react-fast-compare-2.0.4.tgz",
       "integrity": "sha512-suNP+J1VU1MWFKcyt7RtjiSWUjvidmQSlqu+eHslq+342xCbGTYmC0mEhPCOHxlW0CywylOC1u2DFAT+bv4dBw=="
     },
+    "react-ffmpeg": {
+      "version": "0.0.35",
+      "resolved": "https://registry.npmjs.org/react-ffmpeg/-/react-ffmpeg-0.0.35.tgz",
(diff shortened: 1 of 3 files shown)


NdibeRaymond:
should we be doing video compression on the frontend? we already have ffmpeg on the backend so it makes sense to use the backend installation for that. I do have a concern about ffmpeg though. The last time I attempted this the compression took a long time and the compression results wasn't that appealing (this might be because I was trying to compress a large video file with a loss-less format though). If you find an optimal ffmeg setting that allows for a quick compression and great compression result that'd be a huge win.

tuxology:
@gracewzhang Another comment regarding compression: I generally think we should be ok in using a lossy format for now and @NdibeRaymond 's suggestion to do it in the backed also makes sense since this is standard way to do this. 

- For the mobile app, we will find a way to do some clever compression system on the phone later on and depending on bandwidth/scenarios. Currently, Youtube uses VP9/VP8 codec for compression and the containers they use are WebM. They also use mp4 and flv containers and a few other codecs depending on how popular the video is (for each video, they generate multiple encoded videos for different resolutions and deliver them to users depending on their connection speed). We don't have to be this complex at this stage. We can just select a lossy, fast compression speed , variable bit rate and low file size format that more-or-less looks _OKish_ on mobile (where our videos will be consumed and uploaded the most). 
- We can also se what PeerTube does:  https://github.com/Chocobozzz/PeerTube They are spiritual pioneers in open source youtube alternatives. 
- We might also want to have some job-queue system in backend for encoding with fail-safes so we can abandon jobs if the system gets overloaded.  Again,. current requirements are not to built another Youtube/PeerTube but to prove that we can encode stuff in the backend and show results thus saving space for storage as well as data during usage.

ghost:
Moved to a new branch

PR #351, opened 2022-03-29
Add notification grouping code

### LINKED ISSUE #333: Notification Grouping
## Description

Based on Cindy's designs we want to group notifications that are triggered by the same event and happen in a certain time interval. 

## Learning objectives

- Frontend Design
- Django REST Framework

## Todo

- [ ] Update get endpoint to group notifications based on their type/date added
- [ ] Update frontend to support showing grouped notifications (2 profile pics)

## Useful links
- https://www.figma.com/file/jRQdwIwXdgmXEe5L0XLgno/ZubHub?node-id=0%3A1

AndrewLester:
## Summary

Closes #333 

## Changes

- When creating new notifications, if a notification is valid for grouping, check for more recent similar notifications and "group" them
- Use many-to-many field for notification sources

## Requests / Responses

**Response**
image


zubhub_backend/zubhub/notifications/utils.py
```diff
@@ -1,5 +1,46 @@
...
+    prev_notifications = []
+    if notification_type in grouped_notification_types:
+        try:
+            filters = {
+                'recipient': recipient,
+                'type': notification_type
+            }
+            if check_link:
+                filters['link'] = check_link
+            prev_notifications = Notification.objects.filter(**filters).order_by('-date')[:30]
```
ghost: Instead of getting the 30 most recent notifications we could filter based on all notifications within the past hour (or whatever the ```recent_notification_time``` is). I think this would also prevent the need for the if statement on line 27.

ghost: If we go with the idea above, it could be kinda confusing that we're working with an array when we should get at most one field that matches this criteria. To make it a little clearer, maybe we could use [this](https://docs.djangoproject.com/en/dev/ref/models/querysets/#first).

zubhub_backend/zubhub/creators/utils.py
```diff
@@ -290,11 +290,12 @@ def send_notification(users: List[Creator], source: Creator, contexts,
...
             sms_contexts.append(context)
 
+        template_name = get_notification_template_name(Setting.WEB, notification_type)
         message = render_to_string(
-            get_notification_template_name(Setting.WEB, notification_type),
+            template_name,
             context
         ).strip()
-        push_notification(user, source, notification_type, message, link)
+        push_notification(user, source, notification_type, message, link, template_name)
```
ghost: I think we briefly talked about this today, but if grouped notifications end up requiring more context than just the ```sources``` we would have to pass ```context``` into ```push_notification``` and then it wouldn't really make sense to keep this web templating stuff here. Since we might end up moving it into ```push_notification``` anyway and it seems cleaner to isolate the web stuff there, it might be a good idea to just do that now.

ghost (commented):
Great work on the PR! This is definitely a lot cleaner than the approach I was initially thinking. I left a few minor comments but otherwise it looks good. Thanks for getting this done so quickly! I still need to do a bit more testing so I'll approve when done with that.

PR #356, opened 2022-03-30
Make "Tinkering Resources" footer link editable from admin backend

### LINKED ISSUE #345: Make "Tinkering Resources" footer link editable from admin backend
The "Tinkering Resources" resources footer is a static link in the code to `http://kriti.unstructured.studio/` We need to make this link editable from the Django admin interface. Here is the location where you can add make this editable. Here a re few tasks that need to be done to achieve this:

- [ ] Create a Footer section in the following are in Django Admin

image

- [ ] In the footer section, create a editable setting which will change the URL. It would look something like this

image

khabdrick:
## Summary
I created a model for Tinkering Resources, then displayed it in the admin so that it can be edited from there.
Closes #345 

## Changes

- zubhub_backend/zubhub/admin.py
- zubhub_backend/zubhub/models.py
- zubhub_backend/zubhub/serializers.py
## Screenshots

I think the `<p>` issue is a general issue because it also shows up in the admin FAQ display.


tuxology:
Thank you @khabdrick :handshake: We will review this in a few days and then merge. Currently waiting for a few of our own PRs to get merged.

NdibeRaymond:
hello @khabdrick , thanks for working on this issue. There are a few things to note here:
* the tinkering resource as pointed out in the issue is actually located in two places on the frontend, the hero section, and the footer. Judging from that, it shouldn't be a standalone model on it's own but rather it should be a part of the Hero model.
* When the `tinkering_resource_url` field is added to the Hero model, you must also ensure it is included in the hero model serializer and also the view responsible to ensure that it is fetched alongside other fields in the Hero model when an API call to fetch the hero model is made.
* When you are done with the above steps, you then need to make sure to refactor the necessary files on the frontend to get the whole thing connected.

NdibeRaymond:
Also, it is important that you indicate interest on an issue and get the issue assigned to you before actually getting started working on a fix for the issue. This ensures that there is no conflict deciding who was assigned the responsibility of fixing the issue.

NdibeRaymond (changes requested):
requesting changes. Read the previous comments for context

khabdrick:
> Also, it is important that you indicate interest on an issue and get the issue assigned to you before actually getting started working on a fix for the issue. This ensures that there is no conflict deciding who was assigned the responsibility of fixing the issue.

I will start working on the changes, can this issue be assigned to me now?

khabdrick:
hello @NdibeRaymond I'm having trouble finding where the hero section is, in the frontend. Can you please help?
I mean like the route I can use to open it on the browser

Issue #344, opened 2022-03-27
Use same border radius for Next and Prev buttons


tuxology:
In paginated view when Prev and Next buttons are present together, their right and left border radius are zero. We need to keep consistent button styling as all other buttons. Restore the border radius for a more consistent UI experience. 

![image](https://user-images.githubusercontent.com/3373857/160296554-eeb48a8a-d454-404b-9132-c55aaaae797f.png)

Deepanshu039:
@tuxology i would like to work on this.

Deepanshu039:
@tuxology i am still working on the issue.

Morai-C:
please if this is still open i would like to work on it

Deepanshu039:
@Morai-C this issue is under process and i am about to make a PR as well in sometime so i suggest you to do look for some other issues. thank you

Morai-C:
> @Morai-C this issue is under process and i am about to make a PR as well in sometime so i suggest you to do look for some other issues. thank you

okay thank you

PR #370, opened 2022-04-02
Corrected border radius for next and prev buttons

Deepanshu039:
## Summary

Use same border radius for Next and Prev buttons #344

In paginated view when Prev and Next buttons are present together, their right and left border radius are zero. We need to keep consistent button styling as all other buttons. Restore the border radius for a more consistent UI experience.

### DIFF
--- a/zubhub_frontend/zubhub/src/assets/js/styles/components/button/buttonStyles.js
+++ b/zubhub_frontend/zubhub/src/assets/js/styles/components/button/buttonStyles.js
@@ -1,7 +1,7 @@
 const styles = theme => ({
   primaryButtonStyle: {
     backgroundColor: '#00B8C4',
-    borderRadius: '30 !important',
+    borderRadius: '30px !important',
     color: 'white',
     '&:hover': {
       backgroundColor: '#03848C',
--- a/zubhub_frontend/zubhub/src/assets/js/styles/components/button/buttonStyles.js
+++ b/zubhub_frontend/zubhub/src/assets/js/styles/components/button/buttonStyles.js
@@ -1,7 +1,7 @@
 const styles = theme => ({
   primaryButtonStyle: {
     backgroundColor: '#00B8C4',
-    borderRadius: 30,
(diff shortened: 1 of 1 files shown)


zubhub_frontend/zubhub/src/assets/js/styles/components/button/buttonStyles.js
```diff
@@ -1,7 +1,7 @@
 const styles = theme => ({
   primaryButtonStyle: {
     backgroundColor: '#00B8C4',
-    borderRadius: 30,
+    borderRadius: '30 !important',
```
kamthamc: We should try to avoid `!important` as much as possible.  And if you’re trying to pass a string you need to pass the units I believe like `30px`

NdibeRaymond:
@Deepanshu039 thank you for working on this issue!
Can you also make your PR up to date with master?

zubhub_frontend/zubhub/src/assets/js/styles/components/button/buttonStyles.js
```diff
@@ -1,7 +1,7 @@
 const styles = theme => ({
   primaryButtonStyle: {
     backgroundColor: '#00B8C4',
-    borderRadius: 30,
+    borderRadius: '30px !important',
```
NdibeRaymond, line 4: you shouldn't be messing with the primaryButtonStyle or anything inside the buttonStyles.js file directly. This is probably better fixed by removing the button group entirely (+ or - some additional changes), not tweaking this file.

NdibeRaymond (changes requested):

zubhub_frontend/zubhub/src/assets/js/styles/components/button/buttonStyles.js
Deepanshu039: sure, @NdibeRaymond point noted and i will make some changes in a way and make a new PR, thanks you so much for your instructions and review.

PR #371, opened 2022-04-03
Video compression

gracewzhang:
## Summary
- Implement video compression

## Changes
- Implement the compress_video task, which takes the original video, compresses it, then deletes the original video so that the video displayed on the site is the compressed version
- Add ffmpy to celery worker


ghost:
@NdibeRaymond Based on your suggestion, we're planning on moving video compression to the backend. When @gracewzhang tried it using just ```ffmpy``` in the ```upload_file``` function inside of the media app the frontend was stuck at 100%. Even though the ```ffmpeg``` command was something simple like removing metadata it was still stuck at 100%. Do you know what the issue might be there? Also, to try to deal with this we're trying to use celery in the media app so that it can compress the video in the background and still let the rest of the project upload. Since celery isn't installed in the media app currently should we go ahead and do that or is there a better location to use it? Adding celery seems straightforward for the media app but I'm not sure how to adjust the ```celery_container``` docker-related files to support it.

tuxology:
@ajain1921 @gracewzhang Just updating here that we offline decided to create a feature flag and allow enabling/disabling the compression feature from the ZubHub Django Admin. Let us know when this is also near completion. Thanks :pray:

PR #385, opened 2022-04-07
Update lockfile version without breaking npm ci

AndrewLester:
## Summary
Update lockfile version without breaking npm ci


NdibeRaymond:
perhaps @AndrewLester  this is more suited for when we try fixing all the dependabot issues? there are also certain packages that needs to be updated so updating all package files at once makes more sense no?

NdibeRaymond:
perhaps @AndrewLester  this is more suited for when we try fixing all the dependabot issues? there are also certain packages that needs to be updated so updating all package files at once makes more sense no?

AndrewLester:
> perhaps @AndrewLester  this is more suited for when we try fixing all the dependabot issues? there are also certain packages that needs to be updated so updating all package files at once makes more sense no? 

This update is simply to fix the outdated lock file version. The reason I did this is so that when people need to install dependencies in the future (I will soon), the package-lock won't be wholly changed.

We can certainly update depdencies another time too, but I just want that lock file version up. Otherwise, my npm automatically updates it and sometimes breaks `npm ci` on the github action without notice.

NdibeRaymond:
I have opinions about changing package-lock.json this way perhaps because we are still going to update package.json to fix some vulnerabilities which in turn will make new changes to package-lock.json again. I'd prefer we do it all at once, but if everyone else think it's ok, we can merge it. @tuxology @kamthamc  waiting for your opinion on this.

kamthamc:
Updating just package-lock might not work? Since its incompatible with the version of npm we are using in docker. I think We ahould update both deps and node at the same time if possible

AndrewLester:
> Updating just package-lock might not work? Since its incompatible with the version of npm we are using in docker. I think We ahould update both deps and node at the same time if possible

I agree that upgrading node is a good idea too, but just to clarify I did test this package-lock in docker and it was successfully installed.

Do you have any preferred node version?

tuxology:
Maybe we ensure that the version used in docker is the preferred version. This way there will be a good parity between production and development? I would say, lets merge this still.

AndrewLester:
This is good to close now since my other PR incorporated it.

PR #391, opened 2022-04-08
Make "Tinkering Resources" footer link editable and attach to frontend

### LINKED ISSUE #345: Make "Tinkering Resources" footer link editable from admin backend

khabdrick:
## Summary

Added the `tinkering_resource_url` to Hero model and also altered the frontend to use the  `tinkering_resource_url`.
Closes #345

## Changes
- Added a `tinkering_resource_url` field to the Hero model, serializer and view.
- I made tinkering resources in the footer point to `kriti.unstructured.studio` if the tinkering_resource_url is empty.
- Modified how explore_ideas_url is used in the hero section to  explore_ideas_url --- if not present ---> tinkering_resource_url --- if not present ----> https://kriti.unstructured.studio

Please pardon the numerous commits, I messed up with some git commands. It won't happen again.


zubhub_frontend/zubhub/src/views/PageWrapper.jsx
```diff
@@ -113,6 +113,8 @@ function PageWrapper(props) {
   const { anchor_el, loading, open_search_form } = state;
   const { t } = props;
   const { zubhub } = props.projects;
```
NdibeRaymond: const { zubhub, hero } = props.projects;

NdibeRaymond (changes requested):
This is nothing major, just a minor change

NdibeRaymond (approved):

Issue #347, opened 2022-03-28
Generate improved and conscise ZubHub documentation


tuxology:
Current developer  documentation for ZubHub resides in multiple places  

- https://github.com/unstructuredstudio/zubhub/tree/master/zubhub_backend/zubhub/docs as well as 
- https://github.com/unstructuredstudio/zubhub#development
- https://github.com/unstructuredstudio/zubhub/tree/master/single_vm_deployment

We currently need a way to bring all the documentation in one place which is concise and easy to use. New documentation should make lives of developers and newcomers to the project more easy than it is now. We have decided for now to being working on the existing markdown documentation we have and integrate it into a statically generated doc using `mkdocs`. We will be using the mkdocs-material project (https://squidfunk.github.io/mkdocs-material/) to build this new documentation. The tasks involved in this work are as follows:

- [x] Consolidate all markdown files and create a proper documentation structure. Get inspired from https://beta.reactjs.org/learn and https://squidfunk.github.io/mkdocs-material/reference/ themselves
- [x] Convert all the documentation in structure the mkdocs understands and ensure that the documentation builds
- [x] Use current ERD diagram generator mechanism as part of the build
- [x] Create a Github Action to build the docs and push them to github pages
- [x] Rename github action to Documentation Deployment and ensure deployment works. 
- [x] Add scripts to copy relevant files to docs and then use them instead of maintaining two copies of files. Refer https://github.com/unstructuredstudio/zubhub/pull/392#issuecomment-1100777444

qxprakash:
Hi @tuxology I would like to work on this issue as an Outreachy 22 applicant please assign me this issue. Thankyou!!

Vaishnavi-Markunde:
Hi, I would like to build tutorial documentation for installation and setting up the environment on Linux OS for extreme beginners. Please assign it to me. Thank you!

tuxology:
@Prakashh21 and @Vaishnavi-Markunde I am assigning it to you both as you requested this :smile_cat: This is a good opportunity to team up and work as well. I would suggest working on first two tasks in the beginning and see how it works out for you. We can begin by creating a `docs` directory in the top level of `zubhub` project and start adding stuff there. Please add a README.md in `docs` as well so we can see an outline of docs and how they are going to be built.

Vaishnavi-Markunde:
Thank you sir, we'll work on it

[...]

PR #392, opened 2022-04-08
Generated documentation 

Vaishnavi-Markunde:
Issue #347 

- Collected all doc files at one place 
- Used mkdocs to create precise documentation for installation and environment setup


tuxology:
Thanks @Vaishnavi-Markunde for the doc setup! We'll review this and possibly set this up for merging quite soon :+1:

NdibeRaymond:
Thanks for submitting this PR @Vaishnavi-Markunde !
There are changes that needs to be done here that I think I have to do. First is, duplicating the docs to multiple folders is not a good idea. With time the different duplicates will drift away from each other. To get it right, we need to create scripts that manage the duplication and delete them when they are no longer needed. I will merge this, then submit a PR to fix the issues.

PR #395, opened 2022-04-11
Working on issue 201 : Allow picking multiple categories

### LINKED ISSUE #201: Improve the UX around categories
It is not immediately clear right now which category option one should pick for their project. Instead of a dropdown, I wonder if we show a grid of options in the form of grayed-out buttons with images illustrating each category, which might be more helpful for the user. It might also be beneficial to seek a UX designer's input on whether or not we should allow picking multiple categories.

alicendeh:
## Summary
Working on issue 201 entitled Allow picking multiple categories and add more category options

Description of PR here...
 Enabling users to add multiple categories at once before submission.
 Closes #201 

## Changes

-Added packages @emotion/react,@emotion/styled,@mui/material 
- Changed select to Stack from @mui/material/Stack material UI library

## Screenshot

Screenshot from 2022-04-11 14-28-29

### DIFF
--- a/zubhub_frontend/zubhub/package-lock.json
+++ b/zubhub_frontend/zubhub/package-lock.json
@@ -1274,10 +1274,131 @@
         "@date-io/core": "^1.3.13"
       }
     },
+    "@emotion/babel-plugin": {
+      "version": "11.9.2",
+      "resolved": "https://registry.npmjs.org/@emotion/babel-plugin/-/babel-plugin-11.9.2.tgz",
+      "integrity": "sha512-Pr/7HGH6H6yKgnVFNEj2MVlreu3ADqftqjqwUvDy/OJzKFgxKeTQ+eeUf20FOTuHVkDON2iNa25rAXVYtWJCjw==",
+      "requires": {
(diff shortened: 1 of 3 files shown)


alicendeh:
Hi @NdibeRaymond  and @tuxology  could you please take a look at this.
Also, i have a question

 Does the API allows the submission of an array of categories

tuxology:
Currently these categories are hard-coded and fixed. We came up with around 10 of them based on project types. "Tags" however are dynamic. There is a task related to reorganizing categories and tags. Tags are intended to be uses as hashtags etc, while categories are more rigid and are good for browsing using "breadcrumbs"  style navigation. We don't have that support currently as well and this work is still in ideation phase (check https://github.com/unstructuredstudio/zubhub/wiki/Feature-Roadmap-&-Ideas-2022#35-tags-and-categories-restructure). In such case, this PR does not really help in proving the UX for now. If you want you can propose a new idea of Category vs Tags resolution, please start a discussion! For me, it mostly looks like categories == breadcrumbs and tags are what we add as hashtags under description. Therefore, this will also change the UX of individual projects

Issue #343, opened 2022-03-27
Allow basic theme changes to be done from Django Admin


tuxology:
ZubHub currently uses 3 primary colors. There is also a pre-defined font family (Raleway). We want to allow the Zubhub users to change these values from the backend using the admin interface. There are few tasks we need to do to achieve this:

- [ ] Extract colors, fonts and Font sizes being used and convert them to CSS variables. These values are in CSS files as well as embedded in JS.
- [ ] Allow modification of these variables from the backend from ZubHub settings. Add a new item called "Theme" in the area below.

![image](https://user-images.githubusercontent.com/3373857/160296103-04de716b-698d-492b-af96-fec7c72d8d30.png)

untari:
Hi @tuxology  I'm trying to set up the repo in my local machine and I would like to work on these issues. Do I have to be assigned first to be able to work on this issues?

tuxology:
You can begin working on this @untari I'll assign it to you.

untari:
Hi @tuxology thanks! Will start working on it

untari:
Hi @tuxology I figured the current workflow for now:
1. User gets the value in the admin page
2. The backend receives the value and saves it to a variable, should this be saved to the db?
3. When the frontend requests the page, it gets the value from the already set variable and uses it in the style of the page.

Can you please confirm this is the correct workflow, or if you suggest something different?

I find I might need to create new API calls to pass the style values from the backend to the frontend, is this correct or is there a better way to pass those vales?

Thanks!

tuxology:
@untari Yes this is almost correct workflow. By extracting colors and styles, what I meant was carefully refactor the code and make these values configuration driven rather than being hardcoded in the code. Here is what I had in mind:

1. By default, there are color values that are same as the standard ZubHub color palette you see now. We stick to only 3 primary colors and their variations + 2 or 3 fonts. A set of these is called as a theme. The values of the theme are either stored in a configuration file or in DB (or a combination). Absence of values in DB/config file defaults the values of variables to the default palette.

![image](https://user-images.githubusercontent.com/3373857/163679908-5c79c6a9-289e-4757-9490-7287af41c795.png)

2. The fronted uses the colors from this theme
3. A user has a "Theme" section in the admin panel from where they can individually set values in color codes for these variables)
4. When user clicks save in this theme section, the variables values are updated in config/ DB as you suggested
[...]

PR #396, opened 2022-04-11
Allow basic theme changes to be done from Django Admin

untari:
Work in progress for issue #343. **Do not merge yet**.

This PR is here to show my progress and be able to receive feedback on my work.

## Task list
- [ ] Extract colors, fonts and Font sizes being used and convert them to CSS variables. These values are in CSS files as well as embedded in JS. 
- [ ] Allow modification of these variables from the backend from ZubHub settings. Add a new item called "Theme" in the area below (see issue #343).

### First try for one value
- [x] Get familiar with Makefile commands.
- [x] Get familiar with codebase.
- [x] Identify one color variable you can modify.
- [x] Create a Button for it in then admin panel.
- [ ] Add code, so the Button modifies the color value'.

### Extend for all the CSS values
- [ ] Extract colors, fonts and Font sizes being used and convert them to CSS variables.
- [ ] Add them to the "Theme" Item.


zubhub_backend/zubhub/zubhub/urls.py
```diff
@@ -66,6 +66,7 @@
             path('api/help/', HelpAPIView.as_view()),
             path('api/privacy/', PrivacyAPIView.as_view()),
             path('api/faqs/', FAQAPIView.as_view()),
+            path('api/faqs/', ThemeAPIView.as_view()),
```
Ritsyy, line 69: Let's choose a new URL for this

zubhub_backend/zubhub/zubhub/views.py
```diff
@@ -240,6 +240,21 @@ class FAQAPIView(ListAPIView):
...
+    queryset = Theme.objects.all()
+    serializer_class = ThemeSerializer
+    permission_classes = [AllowAny]
+    throttle_classes = [GetUserRateThrottle, SustainedRateThrottle]
+
+    def get_object(self):
+        obj = self.get_queryset()[:1]
+        if obj:
+            return obj[0]
+        return None
```
Ritsyy: we can replace this block with - 
`self.get_queryset().last()`

zubhub_backend/zubhub/zubhub/urls.py
untari: Hi @Ritsyy  I was thinking to name it *api/theme*. What do you think ?

zubhub_backend/zubhub/zubhub/views.py
untari: Hi @Ritsyy I just replaced it. Thanks

PR #403, opened 2022-04-15
updated status code for AccountStatusAPIView

### LINKED ISSUE #401: `AccountStatusAPIView` status code inconsistency
**Describe the bug**
The status code when the user is authenticated is the same as when the user is unauthenticated. I am thinking unauthenticated users should return 404 instead?
**Screenshots**

khabdrick:
## Summary
Update status code in `AccountStatusAPIView` to return 404 when the user doesn't exist. 
Closes #401

### DIFF
--- a/zubhub_backend/zubhub/creators/views.py
+++ b/zubhub_backend/zubhub/creators/views.py
@@ -74,7 +74,7 @@ class AccountStatusAPIView(APIView):
             return Response({'detail': _('Your account is active.')}, status=status.HTTP_200_OK)
         else:
             return Response({
-                'detail': _("Account doesn't exist, is inactive or has been deleted.")}, status=status.HTTP_200_OK)
+                'detail': _("Account doesn't exist, is inactive or has been deleted.")}, status=status.HTTP_404_NOT_FOUND)
 
 
 class UserProfileAPIView(RetrieveAPIView):
--- a/zubhub_backend/zubhub/creators/views.py
+++ b/zubhub_backend/zubhub/creators/views.py
@@ -74,7 +74,7 @@ class AccountStatusAPIView(APIView):
             return Response({'detail': _('Your account is active.')}, status=status.HTTP_200_OK)
         else:
             return Response({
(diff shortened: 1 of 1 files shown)


zubhub_backend/zubhub/creators/views.py
```diff
@@ -74,7 +74,7 @@ def get(self, request, format=None):
             return Response({'detail': _('Your account is active.')}, status=status.HTTP_200_OK)
         else:
             return Response({
-                'detail': _("Account doesn't exist, is inactive or has been deleted.")}, status=status.HTTP_200_OK)
+                'detail': _("Account doesn't exist, is inactive or has been deleted.")}, status=status.HTTP_404_NOT_FOUND)
```
Ritsyy, line 77: I think status code 403 or 401 will be more suitable. Since this is related to access for the user.

khabdrick: Yeah, you are right. I have corrected that.

NdibeRaymond: On the contrary, I believe 404 is the right status code to use here. Here `AccountStatusAPIView` treats user accounts like a resource and is simple trying to verify if the account exists or not. The authentication process here is simply there to help achieve that end goal, so unauthorized users are allowed to access this endpoint. Sorry for the confusion @khabdrick

khabdrick: No problem @NdibeRaymond 
Based on your comment on the pr, should I close this?

NdibeRaymond: I've removed the comment. I saw that you need this for the test is that correct?

khabdrick: Yes, you are correct.

NdibeRaymond: ok this will be merged soon to allow you continue with the test

NdibeRaymond:
changed status code to 404. approving this now

NdibeRaymond (approved):

PR #390, opened 2022-04-08
Fixed issue 211

alicendeh:
## Summary
Realigned buttons in project page and reduced vertical spacing between clab,bookmark,view button row and photos div

## Changes

- Realigned buttons in Project Page
-  Reduced vertical spacing between clab,bookmark,view button row and photos div

## Screenshots

Screenshot from 2022-04-08 05-28-29

PR #409, opened 2022-04-18
fixed uneven padding / spacing between icons on the video's button bar, more visible when the bar becomes horizontal.

### LINKED ISSUE #173: Uneven padding / spacing between icons on the video's button bar
Uneven padding / spacing between icons on the video's button bar, more visible when the bar becomes horizontal. The horizontal bar is full width which is  not ideal as well. It should be grouped on the right with some indication that they are clickable. Before working on this, it may be important to sketch some ideas and gather feedback

Screen Shot 2021-05-25 at 8 38 19 PM

godsakani:
## Summary
Fix #173 
Closes #173 
I changed the display to inline-flex, I gave a flex direction to column and align the items to left using justify-content which help fix the mobile horizontal view. Also for the desktop view I gave a padding-right and also height to align the contents.

## Changes
- projectDetailsstyle.js

## Screenshots
mob
161259516-f4b17142-9f77-48ab-80ab-d89c5ae5671d
tab

### DIFF
--- a/zubhub_frontend/zubhub/src/assets/js/styles/views/project_details/projectDetailsStyles.js
+++ b/zubhub_frontend/zubhub/src/assets/js/styles/views/project_details/projectDetailsStyles.js
@@ -105,16 +105,29 @@ const styles = theme => ({
       height: '3.5em',
       width: '100%',
       flexDirection: 'row',
-      justifyContent: 'flex-start',
+      justifyContent: 'left',
     },
   },
   actionBoxButtonStyle: {
-    margin: '0.5em',
+    color: 'white',
+    margin: '1.0em',
     display: 'flex',
-    maxWidth: '100%',
-    minWidth: 0,
-    flexDirection: 'column',
-    '& span': { display: 'flex', flexDirection: 'column' },
+    maxwidth: '100%',
+    minwidth: '0',
+    textalign: 'center',
+    flexdirection: 'column',
+    justifycontent: 'left',
+    
+    
+    '& span': { 
+      display: 'inline-flex', 
+      flexDirection: 'column',
+      justifyContent:'left',
+      paddingright:'0px',
+      height:'50px',
+      width:'50px',
+      
+    },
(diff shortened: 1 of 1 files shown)


tuxology (changes requested):
@godsakani Thanks :pray: There are conflicts since some styles have been already updated as part of #390. Make sure you keep the styles of the right alignment of horizontal bar and the spacing between the text and icon. The spacing also is very close to what was fixed in #390 so there are many overlaps in this PR. There could be very few improvements still (the left padding is not same as right padding for example I think). 

In addition, we need to change icon order, but this is a larger UX improvement. I recommend you propose better UX than what is already there. Some ideas are: moving the bookmark as a separate button on far right, changing the background of "views" since it is a non-interactive element and its confusing to group them in same container etc. So all this needs some fresh ideas.

godsakani:
> @godsakani Thanks 🙏 There are conflicts since some styles have been already updated as part of #390. Make sure you keep the styles of the right alignment of horizontal bar and the spacing between the text and icon. The spacing also is very close to what was fixed in #390 so there are many overlaps in this PR. There could be very few improvements still (the left padding is not same as right padding for example I think).
> 
> In addition, we need to change icon order, but this is a larger UX improvement. I recommend you propose better UX than what is already there. Some ideas are: moving the bookmark as a separate button on far right, changing the background of "views" since it is a non-interactive element and its confusing to group them in same container etc. So all this needs some fresh ideas.

Ok @tuxology  thanks for the review and correction.

godsakani:
@tuxology 
Am just from browsing the #390 and is exactly the same thing.
please what should i do?

tuxology:
I suggest we close this PR. If you feel there could be new improvements that can be done to existing UI, we can keep it open. Let me know!

godsakani:
> I suggest we close this PR. If you feel there could be new improvements that can be done to existing UI, we can keep it open. Let me know!

@tuxology  Is ok if we close the PR. Because alicendeh  already did the improvement as you requested.

PR #410, opened 2022-04-18
Fix issue #207 : Change login welcome msg color and make it bolder

### LINKED ISSUE #207: Add Login or Sign Up button when user clicks share a project/create project when not logged in
Clicking "Share A project" button when not logged-in shows the following:

image

There is no “sign-in” button. The only way to reach the “login” button is by clicking the hamburger menu. We should have dedicated buttons in such cases saying “Login” and “Sign Up” and also notify user that they are not logged in - would they like to login or sing-up before they get an opportunity to share a project

souad988:
## Summary

Improve the UX on login page by making "Welcome to Zubhub" bolder and Changing it's color.
Closes #207

## Changes

- Increase text Weight add color and text Shadow to login title style class.

## Screenshots

login title yellow

### DIFF
--- a/zubhub_frontend/zubhub/src/assets/js/styles/views/login/loginStyles.js
+++ b/zubhub_frontend/zubhub/src/assets/js/styles/views/login/loginStyles.js
@@ -22,9 +22,11 @@ const styles = theme => ({
     color: 'white',
     padding: '0 30px',
   },
-  ttitleStyle: {
-    fontWeight: 900,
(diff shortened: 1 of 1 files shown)


tuxology:
Hi @souad988 Thanks for your efforts, however the proposed UI style does not match the overall design of the page. I just checked the style and it seems it has been fixed already and is bold enough for now, so I'm closing this one! 

image

PR #414, opened 2022-04-18
Create social sharing buttons component, integrate into ProjectDetails.jsx view

### LINKED ISSUE #384: Social Sharing on Web
## Description

Add social sharing buttons to the project page. 

## Learning objectives

- React
- Sharing APIs

## Todo

- [x] Add social sharing buttons 
- [x] Experiment with sharing APIs to see if you can autofill the sharing platform with text

## Useful links

- https://www.figma.com/file/jRQdwIwXdgmXEe5L0XLgno/ZubHub?node-id=0%3A1

anthonycruzmacedo:
## Summary

Create social sharing buttons capable of creating a Facebook post, Whatsapp message, or copying link to user clipboard, with autofilled message for FB and WA, and with

Closes #384

## Changes
* Create component SocialButtons
* Add SocialButtons to ProjectDetails.jsx

## Screenshots

image


NdibeRaymond:
nice feature to have @saxytony!
Few changes:
I think we can use a "white in transparent background" version of the icons here. The current look is not that great. And also the "share" string doesn't do much here.

NdibeRaymond:
Generally, I am a bit concerned about using packages for changes like this. It does look like something we can do without installing packages. The benefit doesn't outweigh the cost (for me atleast).

anthonycruzmacedo:
> nice feature to have @saxytony! Few changes: I think we can use a "white in transparent background" version of the icons here. The current look is not that great. And also the "share" string doesn't do much here.

@NdibeRaymond I agree, I will get rid of the Share string and look for these kinds of icons so that the display looks more uniform. 

> Generally, I am a bit concerned about using packages for changes like this. It does look like something we can do without installing packages. The benefit doesn't outweigh the cost (for me atleast).

No worries here, the buttons can accomplish the same task without the need to install the react-share package. I will change it so that no packages are necessary.

anthonycruzmacedo:
@NdibeRaymond @ajain1921 Just updated the file to fit the new stylings and not depend on packages for functionality.

zubhub_frontend/zubhub/src/components/social_share_buttons/social_share_buttons.jsx
```diff
@@ -0,0 +1,94 @@
...
+            ahref
+          }
+        >
+          <Avatar
+            className={classes.whatsapp}
+            src={
+              'https://pnggrid.com/wp-content/uploads/2021/05/WhatsApp-logo-png-White.png'
+            }
+          />
+        </Button>
```
ghost: To make this a little cleaner, for all the buttons you can prob use [IconButton](https://v4.mui.com/components/icons/#semantic-svg-icons) instead of `Avatar` (which is generally used for profile pictures) from MUI and the outlined icons for Whatsapp/Facebook that MUI provides. This will also decrease the amount of extra styling you need to add.

zubhub_frontend/zubhub/src/components/social_share_buttons/social_share_buttons.jsx
```diff
@@ -0,0 +1,94 @@
+import React from 'react';
+import { Avatar, Button, ButtonGroup } from '@material-ui/core';
+import { makeStyles } from '@material-ui/core/styles';
+import LinkIcon from '@material-ui/icons/Link';
+// import fbicon from 'fbicon.png';
```
ghost: remove comment

zubhub_frontend/zubhub/package.json
```diff
@@ -6,6 +6,7 @@
     "@date-io/date-fns": "^1.3.13",
     "@material-ui/core": "^4.11.2",
     "@material-ui/icons": "^4.11.2",
+    "@material-ui/lab": "^4.0.0-alpha.61",
```
ghost: You can probably just revert the package.json file to what it is on master because there shouldn't be any changes here.

package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "line-clamp": "^1.0.0"
+  }
+}
```
ghost: Delete file

ghost (commented):
Great work Anthony! Thanks for making the changes quickly. There's just a few more small things you can use to make the PR cleaner that I left comments on. You can also probably just revert the `package.json` and `package-lock.json` files to what they are on master because there aren't any extra packages.

ghost (approved):

Issue #423, opened 2022-04-20
Footer alignment error


singhavs:
## Description
**Problem**
The alignment of Footer is not in centre

![before](https://user-images.githubusercontent.com/67657066/164244520-71529195-d93c-49e6-b462-199d388b85f3.png)

singhavs:
Hey! @NdibeRaymond  can i work on this issue?

tuxology:
@asaayushisingh This looks OK to me :thinking: . Are there other examples in other websites where footer text (if too much) is generally aligned to center in mobile views?  @NdibeRaymond What do you think?

srish:
I am closing the PR and the issue for now as there is no further progress made. Further in my personal opinion concern raised in the issue is not entirely valid or urgent.

PR #431, opened 2022-04-21
Enhancement of Footer

singhavs:
## Summary

Description of PR here...
Closes #443

## Changes
 Enhancement of the footer
## Screenshots
**Before**
before

**After**
after (1)

### DIFF
--- a/zubhub_backend/.env.example
+++ b/zubhub_backend/.env.example
@@ -3,7 +3,7 @@ ENVIRONMENT=production
 DEFAULT_FRONTEND_DOMAIN=<your frontend domain>
 DEFAULT_BACKEND_DOMAIN=<your backend domain>
 DEFAULT_DISPLAY_NAME=ZubHub
-DEFAULT_FRONTEND_PROTOCOL=<frontend protocol i.e http or https>
+DEFAULT_FRONTEND_PROTOCOL=http
 DEFAULT_BACKEND_PROTOCOL=<backend protocol>
 
 SECRET_KEY=<django secret key here>
@@ -15,7 +15,7 @@ DEFAULT_MEDIA_SERVER_DOMAIN=<media server domain>
 
 POSTGRES_NAME=postgres
 POSTGRES_USER=postgres
-POSTGRES_PASSWORD=<random password for postgres>
+POSTGRES_PASSWORD=9493221250jA
 POSTGRES_HOST=db
 
 GF_ADMIN_USER=admin
--- a/zubhub_backend/compose/prometheus.yml
+++ b/zubhub_backend/compose/prometheus.yml
@@ -3,7 +3,7 @@ scrape_configs:
     scrape_interval: 5s
     static_configs:
       # - targets:
-      #   - cadvisor:8080
+      #   - cadvisor:8000
 # - job_name: flower
 #   scrape_interval: 5s
 #   static_configs:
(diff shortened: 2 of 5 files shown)


zubhub_backend/.env.example
```diff
@@ -3,7 +3,7 @@ ENVIRONMENT=production
 DEFAULT_FRONTEND_DOMAIN=<your frontend domain>
 DEFAULT_BACKEND_DOMAIN=<your backend domain>
 DEFAULT_DISPLAY_NAME=ZubHub
-DEFAULT_FRONTEND_PROTOCOL=<frontend protocol i.e http or https>
+DEFAULT_FRONTEND_PROTOCOL=http
```
tuxology, line 6: These changes are non essential to the PR. Remove them

zubhub_backend/.env.example
```diff
@@ -15,7 +15,7 @@ DEFAULT_MEDIA_SERVER_DOMAIN=<media server domain>
 
 POSTGRES_NAME=postgres
 POSTGRES_USER=postgres
-POSTGRES_PASSWORD=<random password for postgres>
+POSTGRES_PASSWORD=9493221250jA
```
tuxology, line 18: Same as above :point_up_2:

zubhub_backend/compose/prometheus.yml
```diff
@@ -3,7 +3,7 @@ scrape_configs:
     scrape_interval: 5s
     static_configs:
       # - targets:
-      #   - cadvisor:8080
+      #   - cadvisor:8000
```
tuxology, line 6: Same as above :point_up_2:

zubhub_frontend/package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
tuxology, line 1: This is not needed. Revert to original state

zubhub_frontend/zubhub/src/assets/js/styles/views/page_wrapper/pageWrapperStyles.js
```diff
@@ -209,7 +209,7 @@ const styles = theme => ({
     backgroundColor: '#dc3545',
     boxSizing: 'border-box',
     width: '100%',
-    textAlign: 'left',
+    textAlign: 'center',
```
tuxology, line 212: This will affect footer for full-width non-mobile cases also and will mess up the UX. Before approaching the PR, can we continue discussion on the related issue to see if this is really needed?

tuxology (changes requested):
We need to discuss the issue a bit more before fixing this. I'll keep the PR open in case we go ahead with this

zubhub_frontend/zubhub/src/assets/js/styles/views/page_wrapper/pageWrapperStyles.js
singhavs: Hey! @tuxology As you can see in non-mobile cases its looking fine only
after-2(1)
Before that the alignment was not in the centre

WhatsApp Image 2022-04-22 at 7 26 20 AM

singhavs:
Hey! @tuxology  As you can see after the change the footer in non-mobile cases is fine
after-2(1)

Before the change the alignment of some heading are not in the centre
WhatsApp Image 2022-04-22 at 7 26 20 AM

tuxology:
@asaayushisingh This type of centering is not intended. The elements under Privacy, About and Help are now center aligned which does not work with other elements in footer. Generally these should be left aligned as they were before so there is no net new change needed. The centering of elements in mobile view can be considered, but we need more concrete examples whether that UX is intended on mobiles or not. Lets discuss on the issue #423

PR #432, opened 2022-04-21
Fix issue #203 : Reformulate long usernames for profile menu

### LINKED ISSUE #203: Long usernames getting cut off from the user menu

souad988:
## Summary

Add condition before displaying username in profile menu, for usernames longer than 14 character cover extra characters in the middle exp : souad988elmansouri -> souad988elm...uri 
Add bubble info to display full username when user hover menu username area
Closes #203

## Changes

- Create function to reformulate long names.
- Add condition before displaying username 
- Add mui `</Tooltip>` component to create bubble info

## Screenshots

spliceLongName14

namesLessThan14

### DIFF
--- a/zubhub_frontend/zubhub/package.json
+++ b/zubhub_frontend/zubhub/package.json
@@ -4,13 +4,10 @@
   "private": true,
   "dependencies": {
     "@date-io/date-fns": "^1.3.13",
-    "@emotion/react": "^11.9.0",
-    "@emotion/styled": "^11.8.1",
     "@material-ui/core": "^4.11.2",
     "@material-ui/icons": "^4.11.2",
     "@material-ui/lab": "^4.0.0-alpha.61",
     "@material-ui/pickers": "^3.2.10",
-    "@mui/material": "^5.6.2",
     "@testing-library/jest-dom": "^5.11.6",
     "@testing-library/react": "^11.1.2",
     "@testing-library/user-event": "^12.2.2",
--- a/zubhub_frontend/zubhub/package.json
+++ b/zubhub_frontend/zubhub/package.json
@@ -4,10 +4,13 @@
   "private": true,
   "dependencies": {
     "@date-io/date-fns": "^1.3.13",
+    "@emotion/react": "^11.9.0",
+    "@emotion/styled": "^11.8.1",
     "@material-ui/core": "^4.11.2",
     "@material-ui/icons": "^4.11.2",
     "@material-ui/lab": "^4.0.0-alpha.61",
     "@material-ui/pickers": "^3.2.10",
+    "@mui/material": "^5.6.2",
     "@testing-library/jest-dom": "^5.11.6",
     "@testing-library/react": "^11.1.2",
     "@testing-library/user-event": "^12.2.2",
--- a/zubhub_frontend/zubhub/src/views/PageWrapper.jsx
+++ b/zubhub_frontend/zubhub/src/views/PageWrapper.jsx
@@ -48,7 +48,6 @@ import {
   handleChangeLanguage,
   handleToggleSearchForm,
   closeSearchFormOrIgnore,
-  sliceLongName,
 } from './pageWrapperScripts';
 
 import {
@@ -521,7 +520,7 @@ function PageWrapper(props) {
                         component="span"
                         className={classes.profileStyle}
                       >
-                        {props.auth.username.length <=14 ? props.auth.username : sliceLongName(props.auth.username) }
+                        {props.auth.username }
                       </Typography>
                       </Tooltip>
                     </MenuItem>
--- a/zubhub_frontend/zubhub/src/views/PageWrapper.jsx
+++ b/zubhub_frontend/zubhub/src/views/PageWrapper.jsx
@@ -36,10 +36,8 @@ import {
   FormGroup,
   InputBase,
   TextField,
-  Button,
+  Tooltip,
 } from '@material-ui/core';
-
-import Tooltip from '@mui/material/Tooltip';
 
 import {
   logout,
@@ -523,7 +521,7 @@ function PageWrapper(props) {
                         component="span"
                         className={classes.profileStyle}
                       >
-                        {props.auth.username.length<=14? props.auth.username : sliceLongName(props.auth.username) }
+                        {props.auth.username.length <=14 ? props.auth.username : sliceLongName(props.auth.username) }
                       </Typography>
                       </Tooltip>
                     </MenuItem>
--- a/zubhub_frontend/zubhub/src/views/PageWrapper.jsx
+++ b/zubhub_frontend/zubhub/src/views/PageWrapper.jsx
@@ -36,7 +36,10 @@ import {
   FormGroup,
   InputBase,
   TextField,
+  Button,
 } from '@material-ui/core';
+
+import Tooltip from '@mui/material/Tooltip';
 
 import {
   logout,
@@ -47,6 +50,7 @@ import {
   handleChangeLanguage,
   handleToggleSearchForm,
   closeSearchFormOrIgnore,
+  sliceLongName,
 } from './pageWrapperScripts';
 
 import {
@@ -512,14 +516,16 @@ function PageWrapper(props) {
                     onClose={e => handleSetState(handleProfileMenuClose(e))}
                   >
                     <MenuItem>
+                    <Tooltip title={props.auth.username} placement="top">
                       <Typography
                         variant="subtitle2"
                         color="textPrimary"
                         component="span"
                         className={classes.profileStyle}
                       >
-                        {props.auth.username}
+                        {props.auth.username.length<=14? props.auth.username : sliceLongName(props.auth.username) }
                       </Typography>
+                      </Tooltip>
                     </MenuItem>
                     <MenuItem>
                       <a className={classes.textDecorationNone} href="/profile">
--- a/zubhub_frontend/zubhub/src/views/pageWrapperScripts.js
+++ b/zubhub_frontend/zubhub/src/views/pageWrapperScripts.js
@@ -98,9 +98,3 @@ export const closeSearchFormOrIgnore = e => {
     handleCloseSearchForm();
   }
 };
-
-export const sliceLongName = (name) => {
-  const arr = name.split('')
-  arr.splice(9, arr.length - 11, '...')
-  return arr.join('')
-}
--- a/zubhub_frontend/zubhub/src/views/pageWrapperScripts.js
+++ b/zubhub_frontend/zubhub/src/views/pageWrapperScripts.js
@@ -98,3 +98,9 @@ export const closeSearchFormOrIgnore = e => {
     handleCloseSearchForm();
   }
 };
+
+export const sliceLongName = (name) => {
+  const arr = name.split('')
+  arr.splice(11,arr.length-14,'...')
+  return arr.join('')
+}
--- a/zubhub_frontend/zubhub/src/views/pageWrapperScripts.js
+++ b/zubhub_frontend/zubhub/src/views/pageWrapperScripts.js
@@ -101,6 +101,6 @@ export const closeSearchFormOrIgnore = e => {
 
 export const sliceLongName = (name) => {
   const arr = name.split('')
-  arr.splice(11,arr.length-14,'...')
+  arr.splice(9, arr.length - 11, '...')
   return arr.join('')
 }
--- a/zubhub_frontend/zubhub/package-lock.json
+++ b/zubhub_frontend/zubhub/package-lock.json
@@ -9,13 +9,10 @@
       "version": "0.1.0",
       "dependencies": {
         "@date-io/date-fns": "^1.3.13",
-        "@emotion/react": "^11.9.0",
-        "@emotion/styled": "^11.8.1",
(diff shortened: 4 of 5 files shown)


zubhub_frontend/zubhub/package.json
```diff
@@ -4,10 +4,13 @@
   "private": true,
   "dependencies": {
     "@date-io/date-fns": "^1.3.13",
+    "@emotion/react": "^11.9.0",
```
tuxology: Why do we need this custom styling package? Can't we use something that is already available and being currently used in ZubHub?

zubhub_frontend/zubhub/package.json
```diff
@@ -4,10 +4,13 @@
...
+    "@emotion/styled": "^11.8.1",
```
tuxology: Same as above :point_up_2:

zubhub_frontend/zubhub/package.json
```diff
@@ -4,10 +4,13 @@
...
     "@material-ui/core": "^4.11.2",
     "@material-ui/icons": "^4.11.2",
     "@material-ui/lab": "^4.0.0-alpha.61",
     "@material-ui/pickers": "^3.2.10",
+    "@mui/material": "^5.6.2",
```
tuxology: we are using `material-ui` and not `mui`. There is tooltip being used already in the project, maybe we use that instead of introducing another new package? This is usually good to keep consistent UI and reduce intermixing of styles. Here is a example of tooltip used currently : https://github.com/unstructuredstudio/zubhub/blob/master/zubhub_frontend/zubhub/src/views/edit_profile/EditProfile.jsx#L25

zubhub_frontend/zubhub/src/views/PageWrapper.jsx
```diff
@@ -36,8 +36,11 @@ import {
   FormGroup,
   InputBase,
   TextField,
+  Button,
 } from '@material-ui/core';
 
+import Tooltip from '@mui/material/Tooltip';
```
tuxology: Same as above :point_up_2: Lets try to use existing packages and styles

zubhub_frontend/zubhub/src/views/PageWrapper.jsx
```diff
@@ -512,14 +516,16 @@ function PageWrapper(props) {
...
                     <MenuItem>
+                    <Tooltip title={props.auth.username} placement="top">
                       <Typography
                         variant="subtitle2"
                         color="textPrimary"
                         component="span"
                         className={classes.profileStyle}
                       >
-                        {props.auth.username}
+                        {props.auth.username.length<=14? props.auth.username : sliceLongName(props.auth.username) }
```
tuxology: Usually its nice to keep space between operators and other tokens for better readability, so this should be `length <= 14 ?`

zubhub_frontend/zubhub/src/views/pageWrapperScripts.js
```diff
@@ -98,3 +98,9 @@ export const closeSearchFormOrIgnore = e => {
     handleCloseSearchForm();
   }
 };
+
+export const sliceLongName = (name) => {
+  const arr = name.split('')
+  arr.splice(11,arr.length-14,'...')
```
tuxology: Same as above :point_up_2: Maybe make it ` arr.splice(11, arr.length - 14, '...')`

tuxology (changes requested):
Thanks for this PR! I've left some feedback!

zubhub_frontend/zubhub/package.json
souad988: ok i understand , i will remove it

zubhub_frontend/zubhub/src/views/PageWrapper.jsx
souad988: ok thanks for mentioning that i will make sure to add spaces

zubhub_frontend/zubhub/src/views/pageWrapperScripts.js
souad988: done

NdibeRaymond:
@tuxology what is the consequences of this if user's username is in other languages and not English? perhaps hindi? I do know that characters and words become quirky and not quite intuitive when internationalization comes into play.
* perhaps splitting certain characters can lead to different meaning?
* or maybe value.length can lie and we end up attempting to split names that fits perfectly into the profile menu?

@souad988 when these things are considered, `value.length` doesn't seem to be the right non-brittle way of fixing this. I think using javascript to get the width of the username container (or something similar, perhaps checking for overflows? ) then compare that to the width of the parent container (profile menu) ensures that the characters can never lie to us regardless of the language. or maybe we should just merge this and stop trying to think about all possible outcomes @tuxology ?

tuxology:
@NdibeRaymond  Actually both the points are indeed valid :exploding_head: @souad988 Can we try to attempt this they way Raymond suggests? If it doesn't work, maybe we merge this for now @NdibeRaymond and when we face issues with other languages and scripts, we will fix it then. We are expecting most users will keep their usernames in English for now.

souad988:
> @tuxology what is the consequences of this if user's username is in other languages and not English? perhaps hindi? I do know that characters and words become quirky and not quite intuitive when internationalization comes into play.
> 
> * perhaps splitting certain characters can lead to different meaning?
> * or maybe value.length can lie and we end up attempting to split names that fits perfectly into the profile menu?
> 
> @souad988 when these things are considered, `value.length` doesn't seem to be the right non-brittle way of fixing this. I think using javascript to get the width of the username container (or something similar, perhaps checking for overflows? ) then compare that to the width of the parent container (profile menu) ensures that the characters can never lie to us regardless of the language. or maybe we should just merge this and stop trying to think about all possible outcomes @tuxology ?

@NdibeRaymond  thanks for the time you spent reviewing this PR,  i dont understand what do you mean by " value.length can lie " would you please clarify with an exemple

souad988:
> @NdibeRaymond Actually both the points are indeed valid 🤯 @souad988 Can we try to attempt this they way Raymond suggests? If it doesn't work, maybe we merge this for now @NdibeRaymond and when we face issues with other languages and scripts, we will fix it then. We are expecting most users will keep their usernames in English for now.

@tuxology  can you explain the idea please i don't understand how length can differ ,  and if i compere containers width and it  overflows, is same function i used valid in this case ?

tuxology:
> @tuxology can you explain the idea please i don't understand how length can differ , and if i compere containers width and it overflows, is same function i used valid in this case ?

I can explain a bit here @souad988. On the computer characters used in different languages can have different length since many languages use different sound symbols and punctuations. Hindi, Arabic, hebrew are an example. For example, a name सृष्टि (Srishti) is 7 character length in English but 6 character length in Hindi when calculated by character count. Even though they represent same information. Similarly سعود is 4 characters, but in English Souad its usually 5 characters. This might cause an issue when someone uses a different language for their name and we try to calculate based on English. Its a very corner case of now, but worth considering. I am not sure how to solve it or if JS has some way to resolve length.

tuxology (approved):
LGTM :rainbow_flag: I generally feel lets merge this one for now, but when we start seeing issues related to characters when people use different languages in their usernames, we will revisit this. I'm approving this for now.

zubhub_frontend/zubhub/src/views/PageWrapper.jsx
```diff
@@ -36,6 +36,8 @@ import {
   FormGroup,
   InputBase,
   TextField,
+  Tooltip,
+  Button,
```
tuxology: Oh I see we added a button why was that?

souad988: 😅 yes sure i will remove it

souad988:
> LGTM 🏳️‍🌈 I generally feel lets merge this one for now, but when we start seeing issues related to characters when people use different languages in their usernames, we will revisit this. I'm approving this for now.

@tuxology  thanks for the approval, i will make more research to understand the issue and to find better possible solutions

NdibeRaymond:
> > @tuxology what is the consequences of this if user's username is in other languages and not English? perhaps hindi? I do know that characters and words become quirky and not quite intuitive when internationalization comes into play.
> > 
> > * perhaps splitting certain characters can lead to different meaning?
> > * or maybe value.length can lie and we end up attempting to split names that fits perfectly into the profile menu?
> > 
> > @souad988 when these things are considered, `value.length` doesn't seem to be the right non-brittle way of fixing this. I think using javascript to get the width of the username container (or something similar, perhaps checking for overflows? ) then compare that to the width of the parent container (profile menu) ensures that the characters can never lie to us regardless of the language. or maybe we should just merge this and stop trying to think about all possible outcomes @tuxology ?
> 
> @NdibeRaymond thanks for the time you spent reviewing this PR, i dont understand what do you mean by " value.length can lie " would you please clarify with an exemple

ok, take for example:

`ओ, हे, अरे, अजी`
`oh hey hey hey hey`

both mean the same thing, but one is physically (and by character length) shorter than the other. 
if you use your function 

```
const sliceLongName = (name) => {
  const arr = name.split('')
  arr.splice(11, arr.length - 14, '...')
  return arr.join('')
}
```
we get:
`oh hey hey ...hey`
`ओ, हे, अरे,...अजी`

but then, `ओ, हे, अरे,...अजी` doesn't have the same meaning as `oh hey hey ...hey`. It means `O, hey, mom,...hey`. The `mom` there is as a result of character splitting. Further more, `ओ, हे, अरे, अजी` (the original phrase) is actually a bit shorter than `ओ, हे, अरे,...अजी` (what we ended up with after the split). Look at both of them stacked on top of the other to really understand what I mean:
`ओ, हे, अरे, अजी`
`ओ, हे, अरे,...अजी`

The implication is that we really had no business splitting the hindi character at all. `ओ, हे, अरे, अजी.length` **lied to us in a way**. Yes, the length is `15`, but it probably would have fitted into the space in the profile menu, and the slice operation only suceeded in making the length longer ( initially, the length was `15`, but after the slice operation, the length became `17`)

other words that has the same weird tendency to morph into new meaning when tampered with are:

`सुचक्र:`
`circle:`

when sliced to become `सु...र:`, the meaning morphs into `Tone:`

There are probably others but these where the ones I could get my hands on.

**Imagine if we mistakenly change a user's perfectly good username to something "explicit" , or an insult. Yea, that doesn't sound so great.**

NdibeRaymond:
something that I think will help to avoid some the issues (especially the apart where words morph into new meanings) is to use line clamp instead. This is a purely css approach and will just end the line with ellipsis when it overflows. They probably handle some of the potential edge cases there.

souad988:
@NdibeRaymond " The implication is that we really had no business splitting the hindi character at all. ओ, हे, अरे, अजी.length lied to us in a way. Yes, the length is 15, but it probably would have fitted into the space in the profile menu, and the slice operation only suceeded in making the length longer ( initially, the length was 15, but after the slice operation, the length became 17) " 

-this one can be fixed by removing the 3 characters we add from the 14 we want to display so the function will always display 14 characters 
`const sliceLongName = (name) => {
  const arr = name.split('')
  arr.splice(9,arr.length-11,'...')
  return arr.join('')
}`

-but semantic issues can not be controlled in this case.

NdibeRaymond:
> @NdibeRaymond " The implication is that we really had no business splitting the hindi character at all. ओ, हे, अरे, अजी.length lied to us in a way. Yes, the length is 15, but it probably would have fitted into the space in the profile menu, and the slice operation only suceeded in making the length longer ( initially, the length was 15, but after the slice operation, the length became 17) "
> 
> -this one can be fixed by removing the 3 characters we add from the 14 we want to display so the function will always display 14 characters `const sliceLongName = (name) => { const arr = name.split('') arr.splice(9,arr.length-11,'...') return arr.join('') }`
> 
> -but semantic issues can not be controlled in this case.

we can just abandon this function and use css line-clamp. never do with javascript what can be done with css. And I suspect the people that designed css line-clamp already took care of such edge cases. 
Here is a css line clamp tutorial. textoverflow can also be used to achieve the same thing https://css-tricks.com/almanac/properties/l/line-clamp/

souad988:
> > @NdibeRaymond " The implication is that we really had no business splitting the hindi character at all. ओ, हे, अरे, अजी.length lied to us in a way. Yes, the length is 15, but it probably would have fitted into the space in the profile menu, and the slice operation only suceeded in making the length longer ( initially, the length was 15, but after the slice operation, the length became 17) "
> > -this one can be fixed by removing the 3 characters we add from the 14 we want to display so the function will always display 14 characters `const sliceLongName = (name) => { const arr = name.split('') arr.splice(9,arr.length-11,'...') return arr.join('') }`
> > -but semantic issues can not be controlled in this case.
> 
> we can just abandon this function and use css line-clamp. never do with javascript what can be done with css. And I suspect the people that designed css line-clamp already took care of such edge cases. Here is a css line clamp tutorial. textoverflow can also be used to achieve the same thing https://css-tricks.com/almanac/properties/l/line-clamp/

@NdibeRaymond  thanks for the clarification, i didn't know that English and other language characters are not represented the same way (unicode vs ascii), so string length can be different from what we expect.  

i will try your suggestion this evening

souad988:
Hi @tuxology  @NdibeRaymond 
i used @NdibeRaymond  suggestion, what do you think?
textOverflow

PR #458, opened 2022-04-28
Recommend Projects View and URL

### LINKED ISSUE #417: Create Recommended Projects Endpoint
## Description

Create an endpoint that will return recommended projects for a certain project. 

## Learning objectives

- Django REST Framework

## Todo

- [ ] Determine the heuristic to use
- [ ] Add endpoint that will use a project ID to generate recommended projects 

## Helpful Links
- Notion Page on Recommendation Heuristic - https://www.notion.so/h4iuiuc/fa3efb7a564348e1acd98c80f621a943?v=da708e7029c04ffa8c334456ca8610e9&p=98ed7386f13c4524b21f9238127cab83

anthonycruzmacedo:
## Summary

Create ProjectRecommendAPIView and endpoint to recommend 3 projects based on:
* first: add those in same category containing first tag, then those in same category containing second tag,  ... 
* second: add those in same category
* third: add most popular projects by likes
* stop in any of these steps once we have retrieved three projects

Closes #417 

## Changes
* ProjectRecommendAPIView in /projects/views.py
* recommend endpdoint in /projects/urls.py
* helper functions recommend_projects and recommend_projects_by_likes in /projects/utils.py


zubhub_backend/zubhub/projects/utils.py
```diff
@@ -310,6 +312,76 @@ def get_published_projects_for_user(user, all):
...
+                    continue
+                # if the jth project's tags include the ith tag, add to projects
+                if i in tags_j:
+                    projects.append(j)
+                # stop once we have added 3 projects (inner loop)
+                if len(projects) == 3:
+                    return projects
+            # stop once we have added 3 projects (outer loop)
+            if len(projects) == 3:
+                return projects
```
ghost, line 371: idt this check is needed. the length should only change after an append

zubhub_backend/zubhub/projects/utils.py
```diff
@@ -310,6 +312,76 @@ def get_published_projects_for_user(user, all):
...
+        for i in tags:
+            # loop through each project in the same category in random order and check if it contains this tag
+            for j in Project.objects.filter(category__name=category).order_by('?'):
+                tags_j = list(j.tags.all())
+                # skip original project and any projects already added to projects
+                if j == project or j in projects:
+                    continue
+                # if the jth project's tags include the ith tag, add to projects
+                if i in tags_j:
+                    projects.append(j)
```
ghost, line 365: In order to make this a little cleaner and prevent redundant checks, we can get all the projects that are `PUBLIC` and that aren't the project that was passed into the function with a filter query at the top. Then all subsequent filters can be done on that. This would remove the need for the 361 check and also ensure we aren't looking at projects that are drafts. Also we can add a filter to 358 that also ensures that the projects have tags that include the `ith` tag.  Check [this](https://docs.djangoproject.com/en/dev/ref/models/querysets/#in) for more info. This would also remove the need for the 364 line.

ghost (commented):
Great work! I really like the randomization idea and the fallback of using the most popular projects! I left a few comments on how we can make this more efficient by making less queries to the DB but overall the algo seems to work.

PR #462, opened 2022-04-30
Issue439:Delete account feature doesn't ask for a password; displays no error message when incorrect username is entered

### LINKED ISSUE #439: Delete account feature doesn't ask for a password; displays no error message when incorrect username is entered
**To Reproduce**

1. Visit your profile page https://zubhub.unstructured.studio/profile
2. Click on the three dots on the top right
3. Click "Delete Account" button that pops up
4. You will notice a dialog that asks for username to confirm deletion. Should we not be asking for password instead?
5. Further, if you try to enter a username that's not yours, nothing happens. You don't event get any error message.

**Expected behavior**
* Ideally, we ask for the password when confirming account deletion. 
* In case an incorrect username is entered, an error message should be displayed.
* Also, there are a few spelling errors in the dialog box. How about we change the text in the dialog to "Are you sure you want to delete your account? You will not be able to undo this action later. If you still want to proceed, type your password into the field below and click DELETE." and change the text of the button to "DELETE".

Screen Shot 2022-04-21 at 5 34 11 PM

alicendeh:
## Summary

Description of PR here...
Closes #439

## Changes

- Added an error popup when the user inputs a wrong username
- modified message on the delete account modal
- Verify both username and password before deleting users account(still in progress)

## Screenshots
Screen Shot 2022-04-30 at 14 10 25
Screen Shot 2022-04-30 at 14 12 34


alicendeh:
Hello @tuxology and @srish  how about this please?

tuxology:
@alicendeh So as per this https://github.com/unstructuredstudio/zubhub/issues/439#issuecomment-1140128595 it seems we should submit user deletion request to backend via the same API, then we do a check in backend and only then we proceed with deletion. A simple comparison should be enough since the current user is already having a valid session - so this is an added layer of security for account deletion. @kamthamc Do you have any other suggestion regarding this flow?

kamthamc:
May be like Facebook
1) we need to send otp to email and confirm. This covers the scenario if session is compromised 
2) schedule deletion (may be  after 3 days) but not delete immediately. And recover the account if the user logs back in 3 days. This also covers the scenario if the session/email is compromised

tuxology:
@kamthamc This is a good idea. But this looks a bit complex at this stage - esp. because the email/SMS system is very ZubHub deployment specific and optional initially. So if an admin decides to not have any e-mail based system, the account holders should at least have control over deletion of their account for privacy reasons without having to reply on admins or any other email based system. Maybe we do this in another iteration and add this to our roadmap?

kamthamc:
If the host decides the integration part. May be they can configure if OTP can be used via email/text/any 2FA. If all of them are disabled we could ask for password. 

2nd part should be fine i believe

Issue #475, opened 2022-05-13
Unable to type in the description field in project creation form on IOS


srish:
The pattern was observed on an IOS device (13 pro with software version 15.3.1). I cannot type in the description field in the project creation form. Even though the field gets the focus upon clicking, I cannot see the cursor and type anything in. There is no issue with the web version, though. Also, tested it on Samsung Galaxy A10S (with Android version 10) and it works fine.
 
<img src="https://user-images.githubusercontent.com/288353/168192725-cc3f9584-5e7a-4543-8957-0617ef5bf464.png" width="200px">

PR #477, opened 2022-05-14
fix issue #475

NdibeRaymond:
This is a bit hard to test because I don't use ios. @tuxology can you help out?


tuxology:
Yes, testing it as we speak! Thanks a lot 🙏

tuxology:
Ok, I can confirm that it works :+1:

tuxology (approved):
:rainbow_flag: LGTM!

PR #482, opened 2022-05-18
Display project uploaded  image preview

### LINKED ISSUE #164: Images preview not available in the "edit project" form
For example, in the form below, next to the image field, it says 2 images added but there is no preview of those images available: 
Screen Shot 2021-05-25 at 5 39 17 PM

Ideally, there are image previews shown in the edit project form.

souad988:
## Summary

Display preview of project images that a user uploads in create project form.
Closes #164

## Changes

- Add Grid container  and Paper components to display uploaded images.
- Make Grid items responsive to screen size.

## Screenshots
preview md
preview xs

### DIFF
--- a/zubhub_frontend/zubhub/public/index.html
+++ b/zubhub_frontend/zubhub/public/index.html
@@ -47,6 +47,7 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
     <title>Zubhub</title>
+    <script src="https://kit.fontawesome.com/58aa8d2b83.js" crossorigin="anonymous"></script>
   </head>
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>
--- a/zubhub_frontend/zubhub/src/assets/js/styles/views/create_project/createProjectStyles.js
+++ b/zubhub_frontend/zubhub/src/assets/js/styles/views/create_project/createProjectStyles.js
@@ -9,6 +9,13 @@ const styles = theme => ({
     background:
       'linear-gradient(to bottom, rgba(255,204,0,1) 0%, rgba(255,229,133,1) 25%, rgba(255,255,255,1) 61%, rgba(255,255,255,1) 100%)',
   },
+  imagePreview:{
+    width: '100%',
+    height: '100%',
+  },
+  imagePreviewContainer:{
+    height: '130px',
+  },
   containerStyle: {
     maxWidth: '600px',
     [theme.breakpoints.up('1600')]: {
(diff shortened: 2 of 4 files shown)


souad988:
Hi @srish  @tuxology 
i hope you are doing well, would you please review my PR, and there is a behavior that i noticed and want to open an issue for, when i click the button to upload an image, it overrides the images i already selected, while it is supposed to append the new images to the uploaded files list,  and we can add delete buttons to the preview in case a user want's to remove an image.

tuxology:
@souad988 Thanks, this is a great feature and will really improve the UX. I tested this a bit more and here is some intial   feedback before I begin full review:

1. For normal desktop width, the image preview sizes look ok. However for larger widths (eg. 1594x2454 that I tested) the images become stretched. See below
 
image

2. For smaller widths such as mobile size same issues exists. See below

image

For resolving these, we should should render image in one style, maybe just a fix 1:1 ratio always with max widths for all responsive  theme "breakpoints" 

3. The styling of images should also follow a bit of consistent styling. I suggest we put images with same round edges as the radius of Description input box int the form and and also have an option to remove them using a Red cross button. (imagine a white  :x:  in a red circle  :red_circle:) placed on top right corner as below

image

It is possible that this action of selectively removing images might not exist, but we can work on it. This also goes in line with the suggestion you have here https://github.com/unstructuredstudio/zubhub/pull/482#issuecomment-1131877028 wherein we can also allow users to append images - thus completing a very good UX feature of selectively removing or adding images.

souad988:
Hi @tuxology  @srish @kamthamc 
would you please review this PR, I made preview images resize with screen size with 1:1 ratio , Added icon and function to remove selectively a preview image.

image preview mobile
image preview remove icon

souad988:
in case you don't want to use fontawsome icons we can replace by `import CloseIcon from '@mui/icons-material/Close';`

srish (changes requested):
@souad988 I tested this feature. Its a really cool one :) Though, I think it has a few issues still that I am sharing below: 
* The images are still getting shrinked. For example, I tried uploading the following images: 
Screen Shot 2022-07-03 at 11 33 10 PM
Notice the ones that have text - second in the first row and last in the second row. And, now compare them with the original ones below. Note that this behavior is seen in desktop (3072 × 1920) as well as mobile resolution.





* Another issue is that after uploading a few images, when I try to add some more, the new images replace the previously uploaded images.

* After the project is created with a few images, when I try to edit it, I don't see the preview. It goes missing.

* When I remove an image, the counter doesn't get updated. 

* Finally, I see that there are bunch of errors on the console, which may have been caused due to other features. Just bringing this up so you can maybe verify and ensure that this feature doesn't contribute to any new errors on the console. 

cc @tuxology

Mara-mma:
Hi @srish @tuxology this is my idea for solving the issue
kite_unstructuredstudio
kite 2_unstructuredstudio

Issue #381, opened 2022-04-06
Notification Bugs


ghost:
## Description

There's a few minor bugs left in the notification implementation to work out.

## Learning objectives
- Django REST Framework
- React

## Todo

- [x] If someone performs an action, undoes it, and redoes it, the backend interprets it as grouped notifications. 
- [x] After clicking a notification the panel should disappear.
- [ ] If you open the mobile notification panel on a project with a video the video shows above the panel. (same applies for floating action buttons)
- [x] Grouped notifications from following someone should redirect to the person who was followed.
- [x] You shouldn't need to refresh the page for notification updates to appear. They should be triggered upon opening the notification panel.

pro-coder-devshekhar:
@ajain1921 Can you please assign me this issue?

AndrewLester:
For the todos:
1. Are we considering removing notifications from actions that are undone as the solution for this?
2. No questions
3. No questions
4. Isn't the person who was followed just yourself?
5. No questions

srish:
@ajain1921 @AndrewLester I recently tried testing the notification feature (with two accounts), and it is not working for me. I tried by leaving a comment on a user's project and I don't see any notification for this activity. Furthermore, when I add a comment, I get an error message: "..Sorry it looks like we have hit a snag, please try again later".

ghost:
@srish I think suchakra was having similar issues with his account before as well but we weren't able to reproduce them with the accounts that we're using. I'm thinking it might have something to do with the fact that the accounts are Staff accounts. Are both the accounts that you tried Staff accounts or is there anything different about them compared to normal accounts that a user might create?

srish:
> @srish I think suchakra was having similar issues with his account before as well but we weren't able to reproduce them with the accounts that we're using. I'm thinking it might have something to do with the fact that the accounts are Staff accounts. Are both the accounts that you tried Staff accounts or is there anything different about them compared to normal accounts that a user might create?

@ajain1921 Yes, I tested with both staff accounts.

ghost:
@srish Ok interesting. I'm guessing this issue is specifically for Staff accounts then, which we didn't test the feature with. We'll try to have a fix ready for it soon.

srish:
[...]

PR #485, opened 2022-05-28
Remove setting table to fix missing setting issues

AndrewLester:
## Summary

Fixes last part of #381


NdibeRaymond:
Hello @AndrewLester thanks for working on this. Perhaps there are better ways of doing this?
Removing the Setting model is not the most desired approach because we intend to make more use of this model (like a settings page on the frontend for user account settings).
This issue is obviously related to Accounts created before the introduction of the Settings model and proper data migration.
The preferred way will be to either create a data migration file to handle this or log into the server and do this manually. I have already handled this so closing this task

PR #503, opened 2022-06-21
Connect Activities component to the store

### LINKED ISSUE #502: Connect activities component to the store.
## Todo

- [x] Connect Activities component to the store.
- [x] Add getActivities method to API.

souad988:
## Summary

Connect Activities component to the store.
Add getActivities method to API.

Closes #502

### DIFF
--- a/zubhub_frontend/zubhub/src/api/api.js
+++ b/zubhub_frontend/zubhub/src/api/api.js
@@ -1,5 +1,5 @@
 import i18next from 'i18next';
-
+let jsonActivities = require('./genericActivities.json');
 /**
  * API class containing all the calls to the backend api endpoints
  */
@@ -851,6 +851,9 @@ class API {
     const body = JSON.stringify({ ...args });
     return this.request({ url, method, token, body }).then(res => res.json());
   };
+  getActivities = () => {
+    return jsonActivities
+  }
 }
 
 export default API;
--- a/zubhub_frontend/zubhub/src/store/actions/activityActions.js
+++ b/zubhub_frontend/zubhub/src/store/actions/activityActions.js
@@ -1,3 +1,6 @@
+import API from  '../../api/api'
+let ActivityAPI = new API()
+
 export const setActivities = activities => {
     return dispatch => {
       dispatch({
@@ -6,3 +9,13 @@ export const setActivities = activities => {
       });
     };
   };
+
+export const getActivities = () => {
+    let activities = ActivityAPI.getActivities()
+    return dispatch => {
+      dispatch({
+        type: 'SET_ACTIVITIES',
+        payload: { all_activities: activities },
+      });
+    };  
+  };  
--- a/zubhub_frontend/zubhub/src/store/actions/activityActions.js
+++ b/zubhub_frontend/zubhub/src/store/actions/activityActions.js
@@ -1,0 +1,8 @@
+export const setActivities = activities => {
+    return dispatch => {
+      dispatch({
+        type: 'SET_ACTIVITIES',
+        payload: { all_activities: activities },
+      });
+    };
+  };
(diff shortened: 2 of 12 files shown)


zubhub_frontend/zubhub/src/api/api.js
```diff
@@ -1,5 +1,5 @@
 import i18next from 'i18next';
-
+let jsonActivities = require('./genericActivities.json');
```
kamthamc, line 2: This needs to be changed to import from require

zubhub_frontend/zubhub/src/store/actions/activityActions.js
```diff
@@ -0,0 +1,21 @@
+import API from  '../../api/api'
+let ActivityAPI = new API()
+
+export const setActivities = activities => {
+    return dispatch => {
+      dispatch({
+        type: 'SET_ACTIVITIES',
```
kamthamc: It’s better to use constants for action types, it avoids the typo issues in reducer/actions

zubhub_frontend/zubhub/src/api/api.js
kamthamc: And I think the activities should come from api otherwise every time user opens the page in new tab or reload the page activities are lost.

@tuxology

zubhub_frontend/zubhub/src/views/activities/activities.jsx
```diff
@@ -0,0 +1,37 @@
...
+     props.getActivities()
+   }, [])
+   
+  return (<div>
+             <h1>activities</h1>
+             {activities? 
+               activities.map(activity => (
+                <h2>{activity.title}</h2>
+               ))
+              :'No activities created yet!!'}
```
kamthamc, line 19: Missing translation?

zubhub_frontend/zubhub/src/views/activities/activities.jsx
```diff
@@ -0,0 +1,37 @@
...
+
+
+function Activities(props) {
+   const activities = props.activities.all_activities
+   useEffect(() => {
+     props.getActivities()
+   }, [])
+   
+  return (<div>
+             <h1>activities</h1>
```
kamthamc, line 14: Missing translation

kamthamc (commented):
LGTM. minor comments

can you please delete activitiesScripts if not used?

zubhub_frontend/zubhub/src/api/api.js
tuxology: Yes, of course. I think this is used for testing @souad988 or maybe as as default?

souad988: > And I think the activities should come from api otherwise every time user opens the page in new tab or reload the page activities are lost.
> 
> @tuxology

Hi @kamthamc , Thanks for taking the time to review.
I'm using this json file just temporarily until creating the backend services for the new activities feature

zubhub_frontend/zubhub/src/store/actions/activityActions.js
souad988: do you mean that i create those constants `const SET_ACTIVITIES='SET_ACTIVITIES'  ...` in the same file?

zubhub_frontend/zubhub/src/views/activities/activities.jsx
souad988: Thanks for mentioning that I will make sure to add it, for this PR the task is just to connect the store to this component

zubhub_frontend/zubhub/src/store/actions/activityActions.js
kamthamc: Yes,
const SET_ACTIVITIES='SET_ACTIVITIES

But in different file since its used in both reducer and actions

PR #505, opened 2022-06-28
User stats enhancement

### LINKED ISSUE #504: Restructure the user's stats on the profile.
## Description
Restructure the user's profile.

## Learning objectives

- React
- CSS

## Todo

- restructure user's stats
- test for different lengths of user name
- make it look good in mobile view

Deepanshu039:
## Summary

#504 
restructure the user's profile.

## Screenshots

**new design**

Screenshot (172)

**with small user_name**

Screenshot (173)

**tablet view**

Screenshot (175)

**mobile view**

Screenshot (174)


zubhub_frontend/zubhub/src/assets/js/styles/views/profile/profileStyles.js
```diff
@@ -54,24 +70,50 @@ const styles = theme => ({
...
   },
   moreInfoStyle: {
-    marginLeft: '0.5em',
-    marginRight: '0.5em',
-    fontWeight: 'bold',
+    display: 'flex',
+    flexDirection: 'column',
+    alignItems: 'center',
+  },
+  moreInfoStyle1: {
```
tuxology, line 106: Maybe give it a more relevant name than `moreInfoStyle1 - such as `moreInfoTitle`?

zubhub_frontend/zubhub/src/assets/js/styles/views/profile/profileStyles.js
```diff
@@ -54,24 +70,50 @@ const styles = theme => ({
...
+    fontWeight: '500',
     fontSize: '0.9rem',
     color: '#00B8C4',
+    paddingLeft: '5px',
+    paddingRight: '5px',
+  },
+  moreInfoStyle2: {
```
tuxology, line 113: This could be `moreInfoCount`?

tuxology (changes requested):
While much of this seems right, the projects, followers and following are very close together. In the mockups the distinction was quite clear. Here they seem too close. 

Mockup:

image

If there is no way to reconcile the distance, we might try to put them in some distinct containers to show some sort of separation? We can experiment with stuff like this as well:

image

zubhub_frontend/zubhub/src/assets/js/styles/views/profile/profileStyles.js
Deepanshu039: noted.

zubhub_frontend/zubhub/src/assets/js/styles/views/profile/profileStyles.js
Deepanshu039: noted.

Deepanshu039: > If there is no way to reconcile the distance, we might try to put them in some distinct containers to show some sort of separation? We can experiment with stuff like this as well:

Also, @tuxology I needed a little help in here, you see there is some padding involve on the Count_values which makes it look a little bit distant from the title. So I was trying to remove the padding but couldn't do much(padding is also not inherited, I tried inspecting it ). Can you suggest something or anything which I might be missing so that it can look more closer and similar to the mockup.

Screenshot (42)


Screenshot (41)

tuxology: Maybe its not padding but `line-height`? maybe for this specific container, can you try to reduce it and see how it behaves?

Deepanshu039: @tuxology you see I remove the line-height here, but still, it is same 
Screenshot (176)_LI

tuxology: @Deepanshu039 OK, lets just proceed with this for now. We can do the design stuff later. I feel there is a lot of more stuff to implement for now such as Badges API, latest user activity (probably via API) etc. These are more pressing issues

tuxology (approved):

PR #518, opened 2022-07-06
Container for badges

### LINKED ISSUE #512: Add badges feature on the user's profile
## Todo

- [x] #513
- [x]  Call the API  and handle the response appropriately
- [x] #517
- [x]  Adding reusable stying component to the badges 
- [x]  Add styling to make the look similar to [figma](https://www.figma.com/file/qwDHghGfIPy9740jLylPwH/figma-mockup_1.0?node-id=0%3A1) design

### LINKED ISSUE #513: Add a container for badges on the user's profile.
Add a badge component(container) on the user's profile which will display the badge tags.

### LINKED ISSUE #515: Create a new file to return the titles of the badges in particular categories to the user's profile.

### LINKED ISSUE #517: Create a new file to return the titles of the badges in particular categories to the user's profile.
Create a separate file that will assign the correct badge_title in a particular category according to the likes_count, views_count, project_count, and comments_count on the user's profile.

Deepanshu039:
## Summary

Container to house badges on user's profile.
Closes #513 

## Screenshot

Screenshot (186)

Screenshot (187)


tuxology:
@Deepanshu039 Is this PR ready for review? I see that the location of Badges section is not exactly as we decided on. If you are still working on the PR, change the PR to Draft mode possibly? Or if you plan to do it in another iteration, that is also fine I guess

Deepanshu039:
> @Deepanshu039 Is this PR ready for review? I see that the location of Badges section is not exactly as we decided on. If you are still working on the PR, change the PR to Draft mode possibly? Or if you plan to do it in another iteration, that is also fine I guess

Yes @tuxology  the last task is yet to be done, even in the issues of the badges I have not checked the last step,
and I am working on it right now. 


> If you are still working on the PR, change the PR to Draft mode possibly?

okay.

Deepanshu039:
hi @tuxology this is the final push for the badges feature from my side, this feature is ready for review : )
(values are hardcoded for a better viewing experience).

## Screenshots

Screenshot (196)

Screenshot (197)

Screenshot (198)

Screenshot (199)

zubhub_frontend/zubhub/src/assets/js/styles/views/profile/profileStyles.js
```diff
@@ -4,47 +4,64 @@ const styles = theme => ({
...
-    height: '100%',
-    paddingTop: '1.5em',
-    paddingBottom: '1.5em',
+    width: '80%',
+    height: '80%',
+    paddingLeft: '1em',
+    paddingRight: '1em',
     '& img': {
-      width: '10em',
+      width: '8em',
```
kamthamc: sizes like width, height should use rem instead of em since they might be affected by font-size.
And its good to use em for paddings
https://www.youtube.com/watch?v=N5wpD9Ov_To

zubhub_frontend/zubhub/src/assets/js/styles/views/profile/profileStyles.js
```diff
@@ -54,33 +71,108 @@ const styles = theme => ({
...
+  },
+  moreInfoTitle: {
+    fontWeight: '600',
     fontSize: '0.9rem',
     color: '#00B8C4',
   },
+  moreInfoCount: {
+    fontWeight: '550',
+    fontSize: '3.5em',
+    color: '#00B8C4',
```
kamthamc, line 124: its better to use colors from theme at all places

zubhub_frontend/zubhub/src/views/profile/DisplayBadges.js
```diff
@@ -0,0 +1,71 @@
+export function allBadgesTitle(projects) {
```
kamthamc: Try using prettier in you editor (if its not default formatter, run format with prettier), it can fix a lot of formatting issues.

zubhub_frontend/zubhub/src/views/profile/DisplayBadges.js
```diff
@@ -0,0 +1,71 @@
...
+    badges.push(title_for_likes);
+
+    if(title_for_views)
+    badges.push(title_for_views);
+
+    return badges;
+}
+
+function addValues(projects){
+    let total_count= {
```
kamthamc: if the constants won't change please use const

zubhub_frontend/zubhub/src/views/profile/DisplayBadges.js
```diff
@@ -0,0 +1,71 @@
...
+        total_count.views+= project.views_count;
+       }
+  
+   return total_count;
+}
+
+
+
+function projectBadgeTitle(count) { 
+    if (count > 100) return 'Expert Builder'
```
kamthamc: these badges needs translations?

zubhub_frontend/zubhub/src/views/profile/Profile.jsx
```diff
@@ -100,11 +102,12 @@ function Profile(props) {
...
       const obj = values[0];
       const drafts = values[1] || {};
+      const badges=allBadgesTitle(obj.results)
 
       if (obj.profile) {
         parseComments(obj.profile.comments);
       }
-      handleSetState({ ...obj, ...drafts });
+      handleSetState({ ...obj, ...drafts, badge_tags:badges });
     });
```
kamthamc: Need to handle catch block. If the API fails or if there is any JS error in the then block. Depending on use case we could show user an error or just console log. If we we are using and client side error tracker it would be helpful

zubhub_frontend/zubhub/src/views/profile/DisplayBadges.js
```diff
@@ -0,0 +1,71 @@
+export function allBadgesTitle(projects) {
+    const {comments, likes, views}= addValues(projects);
+    let badges= [];
+
+    const title_for_projects=projectBadgeTitle(projects.length);
```
kamthamc: Its better to stick to single case either camelCase or snake_case instead of mixing up

zubhub_frontend/zubhub/src/views/profile/Profile.jsx
```diff
@@ -282,34 +315,70 @@ function Profile(props) {
...
+                >
+                  {!profile.members_count
+                    ? t('profile.about.label1')
+                    : t('profile.about.label2')}
+                </Typography>
+                {profile.bio
+                  ? profile.bio
+                  : !profile.members_count
+                  ? t('profile.about.placeholder1')
+                  : t('profile.about.placeholder2')}
```
kamthamc: FYI, it doesn't apply to PR but it  would be easy to understand what the translation might be if we add context in the key.
For example if we add `profile.about.memberPlaceholder` and `profile.about.groupPlaceholder` instead of placeholder1 and placeholder2 it would be easy to know what kind of message will be displayed

zubhub_frontend/zubhub/src/assets/js/styles/views/profile/profileStyles.js
Deepanshu039: noted.

zubhub_frontend/zubhub/src/views/profile/DisplayBadges.js
Deepanshu039: noted, and done.

zubhub_frontend/zubhub/src/views/profile/DisplayBadges.js
Deepanshu039: these are variables that are making count of values inside a loop and adding them, means they are changing that's why I have used let.

zubhub_frontend/zubhub/src/views/profile/DisplayBadges.js
Deepanshu039: Okay, I'll add them in the translation file. working....

zubhub_frontend/zubhub/src/views/profile/Profile.jsx
Deepanshu039: yeah, we should consider that case also. on it

zubhub_frontend/zubhub/src/views/profile/DisplayBadges.js
Deepanshu039: for variables, I have used snake_case and for functions, I have used camelCase so that they can be easily differentiable. Please tell me if it is a good practice or not?
or should I make all of them in camelCase only. @kamthamc

zubhub_frontend/zubhub/src/views/profile/Profile.jsx
Deepanshu039: this was the original translation. Even I was confused about it in the beginning. 
yeah, I'll update that too.

Deepanshu039: thankyou so much @kamthamc for all of your comments, I'll definitely work on them and update you soon.🙂

zubhub_frontend/zubhub/src/assets/js/styles/views/profile/profileStyles.js
Deepanshu039: Yeah, this is necessary to maintain consistency even though I have followed the approach, and the color in the above class is the same as it was in the original code.

zubhub_frontend/zubhub/src/views/profile/DisplayBadges.js
kamthamc: In this case the values inside total_values change but not total_values itself. So const can be used

zubhub_frontend/zubhub/src/views/profile/DisplayBadges.js
kamthamc: In all the JS projects i have worked or seen its all camel case, But its completely up to the team. I am sure if there are any

@tuxology

Deepanshu039: okay, I'll stick to the camelCase only.

Deepanshu039:
hi @kamthamc I have done all of the changes the way you suggested, can you see it now if everything is okay or not.
Thankyou so much for your time....

zubhub_backend/zubhub/creators/admin.py
```diff
@@ -211,10 +219,19 @@ def save_model(self, request, obj, form, change):
         super(CreatorAdmin, self).save_model(
             request, obj, form, change)
 
+class BadgeAdmin(admin.ModelAdmin):
+    list_display = ["badge_title", "created_on", used_by]
+    search_fields = ["badge_title", "created_on"]
+    list_filter = ["created_on"]
+    list_per_page = 50
```
srish, line 226: Are we setting some sort of limit on the number of badges, curious?

zubhub_backend/zubhub/creators/utils.py
```diff
@@ -456,3 +459,118 @@ def send_notification(users: List[Creator], source: Creator, contexts,
 #         EmailAddress.objects.create(
 #             user=user, email=email, primary=False, verified=False
 #         )
+def set_badge_view_category(creator):
+    views_count= Project.objects.filter(creator= creator).aggregate(Sum(('views_count')))["views_count__sum"]
+
+    List = ["Getting Famous", "Person of Interest", "Popular Projects", "Idea Factory"]
```
srish, line 465: Should we move this list out of the code to a central place so it is easily discoverable and modifiable?

zubhub_backend/zubhub/creators/utils.py
```diff
@@ -456,3 +459,118 @@ def send_notification(users: List[Creator], source: Creator, contexts,
...
+    print(views_count, "inside_view_category")
```
srish, line 466: Remove this statement that you may have been using for debugging sake.

zubhub_backend/zubhub/creators/utils.py
```diff
@@ -456,3 +459,118 @@ def send_notification(users: List[Creator], source: Creator, contexts,
...
+
+    if(views_count > 100000):
```
srish, line 468: A lot of what is happening here seems duplicate code. I wonder if there is a creative way to reduce the code here, possibly by using a function that removes and adds badges from the list upon checking if a specific condition is met.

zubhub_backend/zubhub/creators/utils.py
```diff
@@ -456,3 +459,118 @@ def send_notification(users: List[Creator], source: Creator, contexts,
...
+    else:
+        creator.badges.remove(Badge.objects.get(badge_title = List[3]))
+        creator.badges.remove(Badge.objects.get(badge_title = List[2]))
+        creator.badges.remove(Badge.objects.get(badge_title = List[1]))
+        creator.badges.remove(Badge.objects.get(badge_title = List[0]))
+
+def set_badge_like_category(creator):
+    likes_count= Project.objects.filter(creator= creator).aggregate(Sum(('likes_count')))["likes_count__sum"]
+    
+    List= ["Interesting Projects", "Favourite Kid", "Captain Projects"]
```
srish, line 495: Same for this list. I would suggest moving it out from this file to a central place, perhaps a JSON file.

zubhub_backend/zubhub/creators/utils.py
```diff
@@ -456,3 +459,118 @@ def send_notification(users: List[Creator], source: Creator, contexts,
...
+
+    if(likes_count > 1000):
```
srish, line 497: Same for this if possible try to reduce code.

zubhub_backend/zubhub/creators/utils.py
```diff
@@ -456,3 +459,118 @@ def send_notification(users: List[Creator], source: Creator, contexts,
...
+        creator.badges.remove(Badge.objects.get(badge_title = List[2]))
+        creator.badges.remove(Badge.objects.get(badge_title = List[1]))
+        creator.badges.remove(Badge.objects.get(badge_title = List[0]))
+
+
+def set_badge_comment_category(creator):
+    creator_id= creator.id
+    comments_count = Comment.objects.filter(creator__id= creator_id).count()
+
+    List = ["Helping Hand", "Always Available", "Expert Advisor"]
```
srish, line 520: Same goes for this list as well. Take it out.

zubhub_backend/zubhub/creators/utils.py
```diff
@@ -456,3 +459,118 @@ def send_notification(users: List[Creator], source: Creator, contexts,
...
+
+    if(comments_count > 1000):
```
srish, line 522: And, if possible reduce the code here as well.

zubhub_backend/zubhub/creators/utils.py
```diff
@@ -456,3 +459,118 @@ def send_notification(users: List[Creator], source: Creator, contexts,
...
+
+    else:
+        creator.badges.remove(Badge.objects.get(badge_title = List[2]))
+        creator.badges.remove(Badge.objects.get(badge_title = List[1]))
+        creator.badges.remove(Badge.objects.get(badge_title = List[0]))
+
+
+def set_badge_project_category(creator, projects_count):
+    project_count= creator.projects_count
+    if(projects_count < project_count):
```
srish, line 543: This seems like a confusing condition. Consider changing the name of one of the variables, so the two can be easily differentiated.

zubhub_backend/zubhub/creators/utils.py
```diff
@@ -456,3 +459,118 @@ def send_notification(users: List[Creator], source: Creator, contexts,
...
+        creator.badges.remove(Badge.objects.get(badge_title = List[1]))
+
+    else:
+        creator.badges.remove(Badge.objects.get(badge_title = List[2]))
+        creator.badges.remove(Badge.objects.get(badge_title = List[1]))
+        creator.badges.remove(Badge.objects.get(badge_title = List[0]))
+
+
+def set_badge_project_category(creator, projects_count):
+    project_count= creator.projects_count
```
srish, line 542: Leave space between `project_count` and `=`

zubhub_backend/zubhub/creators/utils.py
```diff
@@ -456,3 +459,118 @@ def send_notification(users: List[Creator], source: Creator, contexts,
...
+    if(projects_count < project_count):
+          queryset=Project.objects.filter(creator= creator)
+          if( queryset.exists()):
+                set_badge_view_category(creator)
```
srish, line 546: Seeing these three functions all together makes me wonder if we should just have one function, `set_badge_category`, and pass the type as the parameter.

zubhub_backend/zubhub/creators/utils.py
```diff
@@ -456,3 +459,118 @@ def send_notification(users: List[Creator], source: Creator, contexts,
...
+                set_badge_like_category(creator)
+                set_badge_comment_category(creator)
+
+    List = ["Hatchling", "Flying Bird", "Master of the sky", "Expert Builder"]
```
srish, line 550: Same goes for this, move it out from the code..

srish (changes requested):
I've left a few comments mainly around improving code readability. The majority of comments are for `zubhub_backend/zubhub/creators/utils.py`. These comments are around moving out the badges list for various categories in a central location so they are easily discoverable and modifiable. It is also about investigating if the multiple functions `set_badge_view_category`, `set_badge_comments_category`, and `set_badge_like_category` can be combined into one or, if possible, code reduced in each one of them through the creative use of functions, operators, lists, etc.

zubhub_backend/zubhub/creators/admin.py
Deepanshu039: 50 here means the admin panel where all the badges are listed will show 50 entries on a single page. When the number of badges exceeds 50, they will be on 2nd page

srish: Hmm, so then it probably should be something like `badges_per_page` as that explain the intent of the variable fairly well?

Deepanshu039: list_per_page is an attribute set and here I am just setting its value.

Deepanshu039:
superseded by other PRs.

PR #524, opened 2022-08-14
Badge titles on user's profile

Deepanshu039:
## Summary

Show automated category(total likes, views etc.) badge titles on the user's profile. Custom titles can also be given to a specific user from the admin panel.

## Screenshots

Screenshot (71)

Screenshot (72)


zubhub_backend/zubhub/creators/management/commands/populate_initial_badges.py
```diff
@@ -0,0 +1,17 @@
+from creators.models import Badge
+from django.core.management.base import BaseCommand
+
+
+class Command(BaseCommand):
+    help = 'Populate Badge table with initial badges'
+
+    def handle(self, *args, **kwargs):
+        if Badge.objects.all().count() < 1:
```
siddhism: we can change it to == 0

zubhub_backend/zubhub/creators/utils.py
```diff
@@ -456,3 +458,90 @@ def send_notification(users: List[Creator], source: Creator, contexts,
 #         EmailAddress.objects.create(
 #             user=user, email=email, primary=False, verified=False
 #         )
+
+def set_badge_view_category(creator):
+    views_count= Project.objects.filter(creator= creator).aggregate(Sum(('views_count')))["views_count__sum"]
+
+    List = ["Getting Famous", "Person of Interest", "Popular Projects", "Idea Factory"]
```
siddhism: List is a reserved keyword, can we name it something like badges_list

zubhub_backend/zubhub/creators/utils.py
```diff
@@ -456,3 +458,90 @@ def send_notification(users: List[Creator], source: Creator, contexts,
...
+
+    creator.badges.remove(Badge.objects.get(badge_title = List[3]))
+    creator.badges.remove(Badge.objects.get(badge_title = List[2]))
+    creator.badges.remove(Badge.objects.get(badge_title = List[1]))
+    creator.badges.remove(Badge.objects.get(badge_title = List[0]))
+    
+    if(views_count > 100000):
```
siddhism: can we move these constants to a hashmap. If this logic is used at multiple places it'll be helpful, or otherwise as well.
something like
`getting_famous: 10000` and pick the 10000 from this map.

zubhub_backend/zubhub/creators/utils.py
```diff
@@ -456,3 +458,90 @@ def send_notification(users: List[Creator], source: Creator, contexts,
...
+    
+def set_badge_like_category(creator):
+    likes_count= Project.objects.filter(creator= creator).aggregate(Sum(('likes_count')))["likes_count__sum"]
+    List= ["Interesting Projects", "Favourite Kid", "Captain Projects"]
+
+    creator.badges.remove(Badge.objects.get(badge_title = List[2]))
+    creator.badges.remove(Badge.objects.get(badge_title = List[1]))
+    creator.badges.remove(Badge.objects.get(badge_title = List[0]))
+
+    if(likes_count > 1000):
```
siddhism: same comment for counts check like above. or try to see if there is a better way to reduce if/else blocks with a single function/constant/configuration like constant which returns badge_title based on the count.
We can do this for all 4 type of count checks in this file.

tuxology (approved):
Reviewed and tested on staging :rainbow_flag:

PR #535, opened 2022-10-12
Set up ambassadors page

srish:
These changes include a frontend view for the ambassadors' page and introduces a model in Django to accept information via `summernote` editor from the backend/admin panel.


srish:
@tuxology Could you help review and merge this?

NdibeRaymond:
@tuxology I fixed the issue we discussed on zulip. Though it still needs some css changes. While looking at the new ambassadors page feature, I found out that ambassadors field in the Ambassadors model is a text field. If ambassadors are expected to be users with accounts on the platform, who may or may not have created projects, isn't it better that we use relationships here to link it to the users accounts? that way the projects field in the Ambassadors model will no longer be needed as we can always get the projects through the relationship `user.projects`. I might also be wrong though since I don't have the full context of this new feature

zubhub_backend/zubhub/zubhub/migrations/0007_ambassadors.py
```diff
@@ -0,0 +1,25 @@
+# Generated by Django 3.2 on 2022-10-04 00:47
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
NdibeRaymond, line 6: I removed the changes that was directly added to this migration. since the 0007_ambassadors.py migration has been applied in the past, making more changes to will lead to migration conflict. All the new changes can instead be found in 0008_ambassadors_projects.py migration

zubhub_backend/zubhub/projects/migrations/0007_tag_tag_name_gin_idx.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.2 on 2022-11-07 04:01
```
NdibeRaymond, line 1: We remove this migration in the past because of the error it was causing in production. This error was due to the changes made to table index by Andrew (it is somehow conflicting with the name of an already existing index). Here we had to add it back because we can no longer avoid it. while we can continue iterating on this feature in our local dev environments, we can't merge this PR to master without fixing the bug that this migration file introduces and attempting to do so WILL CAUSE AN ERROR IN PRODUCTION.

tuxology (approved):

zubhub_backend/zubhub/projects/migrations/0007_tag_tag_name_gin_idx.py
tuxology: This is now resolved after @NdibeRaymond 's help in production! :pray:

PR #536, opened 2022-10-20
reverted and pegged django_summernote version to 0.8.11.6

NdibeRaymond:
the bug in #535 is because there was a change in `django_summernote` version and we were using `>=` in the `requirements.txt`.
strictly pinning it to version `0.8.11.6` for now.


NdibeRaymond:
@tuxology and @srish  this is ok to be merged. merging it then pulling the changes into #535 should fix the problem. I made it a separate patch so it's obvious what was done. I'm leaving the merging up to you

tuxology (approved):

Issue #177, opened 2021-05-26
Improvements to search bar


srish:
- [ ] Use darker colored search bar

<img width="325" alt="Screen Shot 2021-05-25 at 8 58 55 PM" src="https://user-images.githubusercontent.com/288353/119600524-13394400-bd9c-11eb-8b65-12e7a629cdaf.png">

- [ ] Center the searchbar on large screens. 
<img width="324" alt="Screen Shot 2021-05-25 at 8 59 02 PM" src="https://user-images.githubusercontent.com/288353/119600533-17656180-bd9c-11eb-951d-d0014f48443e.png">

- [ ] Search bar appears funny in mobile view
<img width="324" alt="Screen Shot 2021-05-25 at 8 59 02 PM" src="https://user-images.githubusercontent.com/288353/167981144-088626bd-393c-4483-bc1c-61e08c8606cc.PNG">

sunnyside88:
Hi there, can I work on this issue?

Jumeb:
@tuxology, I would love to work on this issue. Please can you assign it to me

Jumeb:
> @tuxology, I would love to work on this issue. Please can you assign it to me

Screenshots

![Screenshot from 2022-03-31 22-33-25](https://user-images.githubusercontent.com/38510556/161153332-a07953b3-ec49-4ae5-9787-63907343d027.png)
dark theme applied

![Screenshot from 2022-03-31 22-33-15](https://user-images.githubusercontent.com/38510556/161153475-6339b546-4b69-4550-90b0-0f1a34191942.png)
Centralized search bar

![Screenshot from 2022-03-31 22-35-00](https://user-images.githubusercontent.com/38510556/161153544-87c91ee1-3733-4980-bbd2-0cf8ca9e24ed.png)
hover effect

I would love to make a pull request @srish  and @tuxology

tuxology:
If the PR is ready, you can send it @Jumeb Can you adjust he color in 3rd screenshot to be not so dark? Maybe the selected one could be made lighter or light yellow (and icon changed to red?) Play around with it and send the PR

Jumeb:
> If the PR is ready, you can send it @Jumeb Can you adjust he color in 3rd screenshot to be not so dark? Maybe the selected one could be made lighter or light yellow (and icon changed to red?) Play around with it and send the PR

Ok on it

Jumeb:
[...]

PR #584, opened 2023-03-07
improved, centralized and darkened the search bar (issue-177)

iamatifmoin:
## Summary

I have darkened the colour of the search bar and centralized it. 
I have also added a hover effect, which darkens the colour when the user hovers over it.
Helps in #177

## Changes

- Centralized the search bar
- Darkened the colour of the search bar
- Added hover effect to the search bar

## Screenshots

darkened search bar
Zubhub

hover effect
Zubhub (1)


zubhub_frontend/zubhub/package.json
```diff
@@ -47,7 +47,7 @@
   },
   "scripts": {
     "analyze": "source-map-explorer 'build/static/js/*.js'",
-    "start": "react-scripts start",
+    "start": "react-scripts --openssl-legacy-provider start",
```
tuxology: Why do we need this?

zubhub_frontend/zubhub/src/views/PageWrapper.jsx
```diff
@@ -752,10 +756,7 @@ function PageWrapper(props) {
                   </InputLabel>
                   <Autocomplete
                     options={options}
-                    defaultValue={
-                      props.location.search &&
-                      getQueryParams(window.location.href).get('q')
-                    }
+                    defaultValue={getQueryParams(window.location.href).get('q')}
```
tuxology: Why remove the props here?

tuxology (commented):
Left a few comments, we can approve one we are comfortable with reasoning

zubhub_frontend/zubhub/package.json
iamatifmoin: using `"start": "react-scripts start"` generated the following error for me:
> Error: error:0308010C:digital envelope routines::unsupported

I looked for solutions [here](https://stackoverflow.com/questions/69692842/error-message-error0308010cdigital-envelope-routinesunsupported) and found that the error can be resolved by using `"start": "react-scripts --openssl-legacy-provider start"`

This change, however, should have been made locally only and should not have been committed.
I will reset the change by making another commit.

zubhub_frontend/zubhub/src/views/PageWrapper.jsx
iamatifmoin: I think I made this change unintentionally while trying to figure out the working of the code. 

I will reset the change in a new commit and make sure to go through the changes I make next time.

tuxology (approved):

iamatifmoin:
@tuxology 
When will the PR be merged? Or are some changes required?

PR #613, opened 2023-03-09
Add navigation button to project preview image

### LINKED ISSUE #435: Navigation buttons on the caraousel images.
Left-right navigation buttons can be added in the carousel mode using which user can easily navigate between images. This functionality reduces the effort to preview images in a project and also quick navigation can be done using left-right button.

Screenshot (88)

two left-right buttons can be attached over the images itself when in PREVIEW mode so that user do not need to press back button again and select new image from there.
Using the buttons over the image next or previous image can be previewed.

psalmuelle:
## Summary
Users can now have access to the next and previous buttons while previewing project images. There is also an explicit cancel button on the top right corner of the images.
If project image is one, only the image is shown as there is no need for a carousel.

Closes #435 

## Changes

- Replaced preview image with slider in the preview dialog
- Changed image dialog style
- Added the cancel button to the dialog

## Screenshots

Screenshot 2023-03-09 103357


tuxology:
Thanks @psalmuelle 🤝  Some suggestions:
 - Can we place the red cross on the image corner itself, something like this? 
image
 - In addition, we should also improve the left/right arrows styling a bit more so that they are clearly visible. I think this can also be achieved by decreasing opacity of the modal background (make it more darker - maybe 70% dark)
 - Another suggestion is to increase the size of the image so it takes up more screen size. The issue we see currently is also that image size shown even after clicking is very small. 

Finally, please test this on responsive views as well and share screenshots. After this, we can star code review.

psalmuelle:
Alright.  I will make changes according to your suggestions.
Thank you for the feedback!

psalmuelle:
@tuxology. I have updated the UI according to your suggestions

- I have changed the cancel button's position to the image's top right corner.

- I have also styled the arrow to look bigger and decreased the background opacity.

- Finally, I increased the size of the image in-preview

Screenshot from a PC
correction

Tablet


Mobile


Is there any other thing that can be improved? 
Thank you!

tuxology:
Nice, looks nice for now, I'll review and send feedback @psalmuelle

psalmuelle:
Alright. Thank you!

zubhub_frontend/zubhub/src/assets/js/styles/views/project_details/projectDetailsStyles.js
```diff
@@ -1,3 +1,5 @@
+import zIndex from "@material-ui/core/styles/zIndex";
```
kamthamc, line 1: Use Prettier plugin to format files, the formatting seems to be off.

zubhub_frontend/zubhub/src/assets/js/styles/views/project_details/projectDetailsStyles.js
```diff
@@ -161,15 +163,32 @@ const styles = theme => ({
...
     alignSelf: 'center',
-    width: '80%',
-    maxWidth: '600px',
+    padding:'2px',
+    width: '85%',
+    maxWidth: '840px',
     height: 'auto',
   },
+  cancelEnlargedImageBtn:{
+   backgroundColor: '#F50057',
```
kamthamc, line 182: Use Theme for the colors

zubhub_frontend/zubhub/src/assets/js/styles/views/project_details/projectDetailsStyles.js
```diff
@@ -161,15 +163,32 @@ const styles = theme => ({
...
+   position: 'absolute',
+   right: '6%',
```
kamthamc, line 184: I think the button should be have a constant size from the right?
cc @tuxology

zubhub_frontend/zubhub/src/assets/js/styles/views/project_details/projectDetailsStyles.js
```diff
@@ -161,15 +163,32 @@ const styles = theme => ({
...
+   top: '0%',
+   width: '34px',
+   height: '34px',
+   borderRadius: '100%',
+   border: '2px solid white',
```
kamthamc, line 189: Use themes for colors

zubhub_frontend/zubhub/src/assets/js/styles/views/project_details/projectDetailsStyles.js
```diff
@@ -369,4 +388,31 @@ export function PrevArrow(props) {
   );
 }
 
+export function EnlargedImgArrow(props) {
+  const { className, style, onClick } = props;
+  return (
+    <div
+      className={className}
+      style={{ ...style,
```
kamthamc, line 396: Better to use class for this and FYI the style/any prop spread should be at the end so we can override styles

zubhub_frontend/zubhub/src/views/project_details/ProjectDetails.jsx
```diff
@@ -491,12 +507,57 @@ function ProjectDetails(props) {
...
+                slidesToScroll={1}
+                className= {classes.enlargedImageStyle}
+                nextArrow = {<EnlargedImgArrow/>}
+                prevArrow = {<EnlargedImgArrow/>}
+
+              >
+                {project.images.map(image => (
+                  <div>
+                    <img
+                      style={{
```
kamthamc, line 549: Use classes instead of inline styles

kamthamc: And images need to have alt text, If the image doesn't have title or description properties add empty text for alt

zubhub_frontend/zubhub/src/views/project_details/ProjectDetails.jsx
```diff
@@ -491,12 +507,57 @@ function ProjectDetails(props) {
...
+         {/* Cancel Button on the Preview Images */}
+          <IconButton
+            className={classes.cancelEnlargedImageBtn}
+            onClick={() =>
+              setState({
+                ...state,
+                open_enlarged_image_dialog: !open_enlarged_image_dialog,
+              })
+            }
+            aria-label="close"
```
kamthamc, line 520: aria labels should also be translated

zubhub_frontend/zubhub/src/assets/js/styles/views/project_details/projectDetailsStyles.js
kamthamc: And use styles instead of inline styles

zubhub_frontend/zubhub/src/views/project_details/ProjectDetails.jsx
psalmuelle: I have made note of all corrections, Ill make changes right away. Thank you @kamthamc

tuxology:
@psalmuelle Are we done with this and is this merge ready or are you working still on suggestions from @kamthamc ?

PR #626, opened 2023-03-10
Fix using svg for Play icon instead of png

### LINKED ISSUE #620: Use of SVG instead of PNG in play icon
## The bug
The png for play icon gets blurred out based on screen resolution so replacing the png to an svg for the same would be a great quick fix

## Screenshots
image

darechan:
## Summary

Initially the `Play Icon` for projects had a low resolution png which was replaced by a svg for the same.
Closes #620 

## Changes

- Play-icon png was replaced by play-icon svg
- All the path referencing the png were altered
- The SVG was styled to fit the project card

## Screenshots

image


tuxology:
@darechan Nice 🎉 Can we add a small shadow to it just to keep it consistent with other shadowy elements like card etc? And also the request about color change when hovering? Just slight hue/shade change can be ok. No need to completely make it red/yellow on hover.

darechan:
@tuxology for sure- that would add in a bit more detail

darechan:
@tuxology 
image

shall I add more blur?

tuxology:
Yes, please see that it is close to the blur/direction as in card

darechan:
> Yes, please see that it is close to the blur/direction as in card

image

Updated please check!

tuxology (approved):

PR #660, opened 2023-03-13
removing width for fixing search bar overflow

iamatifmoin:
@tuxology I have removed the `width:40em` line from `pageWrapperStyles.js`

Helps in #177

### DIFF
--- a/zubhub_frontend/zubhub/src/assets/js/styles/views/page_wrapper/pageWrapperStyles.js
+++ b/zubhub_frontend/zubhub/src/assets/js/styles/views/page_wrapper/pageWrapperStyles.js
@@ -40,7 +40,7 @@ const styles = theme => ({
     '& .search-form-input': {
       height: '2.5em',
       maxWidth: '40em',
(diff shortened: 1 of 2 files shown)


zubhub_frontend/zubhub/src/assets/js/styles/views/page_wrapper/pageWrapperStyles.js
```diff
@@ -40,7 +40,7 @@ const styles = theme => ({
     '& .search-form-input': {
       height: '2.5em',
       maxWidth: '40em',
-      width: '40em',
+      // width: '40em',
```
tuxology:
```suggestion

```

tuxology (approved):
Remove instead of commenting

iamatifmoin:
> Remove instead of commenting

@tuxology Alright. 
Will keep that in mind for further such cases.

Just curious though, was this the problem causing the overflow or was it a combination of other such problems?

PR #667, opened 2023-03-14
Password reset bug fix

### LINKED ISSUE #654: Password reset not checking if user already has an account before validating the request.
**Describe the bug**
When we access the password [reset page ](https://zubhub.unstructured.studio/password-reset), and we add an email address that isn't tied to an existing user in the system, we get a success message from the API saying `Password reset e-mail has been sent`
This should not be the case as password reset emails should only be sent to those that had previously created an account on zubhub.

**To Reproduce**

1. Go to https://zubhub.unstructured.studio/password-reset
2. Enter a creative email that you think doesn't exist on the app 
3. Submit the form
4. See error or check the network tab to see the response of the API

**Expected behavior**
When an email doesn't exist on the platform, the API should indicate it and we can let the user know that the email they used doesn't exist on the platform. This will also help us not to send unnecessary emails to people that don't have accounts.

skynette:
## Summary

When an email doesn't exist on the platform, the API should indicate it and we can let the user know that the email they used doesn't exist on the platform. This will also help us not to send unnecessary emails to people that don't have accounts.
Closes #654 

## Changes
- Added Custom Password Reset Serializer to zubhub.serializers
- Added Configuration for the serializer in settings.py

## Screenshots
image

### DIFF
--- a/zubhub_backend/zubhub/zubhub/serializers.py
+++ b/zubhub_backend/zubhub/zubhub/serializers.py
@@ -4,8 +4,13 @@ from projects.pagination import ProjectNumberPagination
 from projects.utils import get_published_projects_for_user
 from projects.serializers import ProjectSerializer
 from math import ceil
+from django.contrib.auth.forms import PasswordResetForm
+from django.contrib.auth import get_user_model
+from django.conf import settings
+from django.utils.translation import gettext as _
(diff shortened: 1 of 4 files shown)


tuxology:
I am closing this since work is already being done in #674 .Please do not duplicate work. Sync with the developer already on the task. Thanks!

PR #674, opened 2023-03-15
Password reset not checking if user already has an account before validating the request

### LINKED ISSUE #654: Password reset not checking if user already has an account before validating the request.

yaya-mamoudou:
## Summary
When a user tries to reset their password, we check if the provided email is linked to an existing account before sending a success request. Otherwise, an appropriate response is sent to inform the user that an account with the email they provided was not found and they should verify their email and try again

Closes #654 

## Changes

- Created a Custom password reset class that extends the PASSWORD_RESET_SERIALIZER  provided by Django rest framework and added another layer of verification to check if email exists in the DB or else return an error.
-  Cleared the form error when submitting In case there was an error displayed from the previous submission.

## Screenshots
https://www.loom.com/share/4835dd48ada641daab233d093bd60e94


tuxology (commented):
Thanks, I've reviewed this 🎉 No real changes needed, but we would have to wait a bit here since I want to test this on staging first @yaya-mamoudou

yaya-mamoudou:
> Thanks, I've reviewed this 🎉 No real changes needed, but we would have to wait a bit here since I want to test this on staging first @yaya-mamoudou

Alright, @tuxology. Thanks🥳

tuxology (approved):
So I did one test on this, and it seems to work on staging. I will merge this a bit later after all other fixes are done. Thanks 🙏

yaya-mamoudou:
> So I did one test on this, and it seems to work on staging. I will merge this a bit later after all other fixes are done. Thanks 🙏

Thanks, @tuxology. I am so happy this got approved. First time writing python code. This issue forced me to learn the basics of python and Django in 3 days. 🎉🎊

PR #681, opened 2023-03-17
Colour updation of buttons for hover effect

### LINKED ISSUE #629: Make buttons hoverable
Hi! @srish @tuxology , Only "Share a Project" button has hover effect but it will be a better idea if we add hover effect to "Explore ideas" and "Be our Ambassador" buttons.

image

darechan:
## Summary

This change in hex colour makes the `darkDangerButtonStyle` for explore idea page and `customWarningButtonStyle` for Be an ambassador button hoverable.
 
Fixes #629

### DIFF
--- a/zubhub_frontend/zubhub/src/assets/js/styles/components/button/buttonStyles.js
+++ b/zubhub_frontend/zubhub/src/assets/js/styles/components/button/buttonStyles.js
@@ -39,23 +39,23 @@ const styles = theme => ({
       color: 'white',
       backgroundColor: '#9A0036',
     },
-  }, 
+  },
   darkDangerButtonStyle: {
     borderRadius: 30,
     backgroundColor: '#9A0036',
     color: 'white',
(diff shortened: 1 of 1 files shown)


zubhub_frontend/zubhub/src/assets/js/styles/components/button/buttonStyles.js
```diff
@@ -39,23 +39,23 @@ const styles = theme => ({
...
       color: 'white',
-      backgroundColor: '#9A0036',
+      backgroundColor: '#5E0423',
     },
   },
   customWarningButtonStyle: {
     borderRadius: 30,
     backgroundColor: '#FECB00',
-    color: '#896e00',
+    color: 'white', // previous colour #896e00
```
tuxology:
```suggestion
    color: 'white',
```

tuxology (approved):

PR #699, opened 2023-03-24
Added 'Save as Draft' option and removal of 'All registered zubhub users' option

cherish2003:
## Summary
This PR renames the options as per operations
The option Only few zubhub users is renamed to Save as Draft and All registered zubhub users is renamed to Only few zubhub users this changes are done because when we select option All registered ZubHub users it is showing for adding usernames but it is not possible to determine all the ZubHub users by taking as input and while we select option Only few zubhub users it saves the project to drafts so it would be perfect for renaming the options so the issue will be solved

## Changes
- Renaming of options in translation.json file

## Screenshots
Screenshot 2023-03-16 at 3 16 17 PM


cherish2003:
@tuxology this is the new PR with the removal of conflicts of merges that are there in previous PR

tuxology (approved):

PR #705, opened 2023-03-26
tests: add tests for activities backend

### LINKED ISSUE #687: Add tests for Models and Views of the 'activities' Backend.
Tests for the activities backend will help notice breaking changes and thus  prevent disruption of 
the application leading to great user experience.

## screen shot 
Screenshot_20230318-085648

Collins-Omariba:
## Summary

Adding of tests for the activities backend django app
Closes #687 

## Changes

- Deleted tests.py file in activities app
- Added tests folder with test_models, test_views and __init__.py file

## Screenshots
**The permissions implementation in the activities app need changing** since: a regular user can create activities though they are not allowed and a non-staff user can toggle publish activity & also get unpublished activities, hence the reason the corresponding tests fail. 

Screenshot (106)

### DIFF
--- a/zubhub_backend/zubhub/activities/tests.py
+++ b/zubhub_backend/zubhub/activities/tests.py
@@ -1,3 +1,0 @@
-from django.test import TestCase
-
-# Create your tests here.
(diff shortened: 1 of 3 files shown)


nerdyk3 (commented):
Nice to have this PR 🎉 

Observation:
- There are many repeated values/strings used in the files 

solution: 
- You can create a constant.py to define the repeated value
- Try to use [django-factory](https://medium.com/analytics-vidhya/factoryboy-usage-cd0398fd11d2) 

Thank You

tuxology (approved):
These tests have not been run on latest code, but approving this so we can have basic testing of activities backend. Thanks 🙏

PR #714, opened 2023-06-05
Currently, the email gets displayed on the user profile. 

### LINKED ISSUE #713: Currently, the email gets displayed on the user profile. We want to avoid that
**Describe the bug**
Currently, the email gets displayed on the user profile. This is caused because, in place of the username field in the register form, the user entered the email.

**To Reproduce**
1. Go to [this](https://zubhub.unstructured.studio/signup) link 
2. Fill out the form and in place of the username, add your email and submit the form
3. go to your profile and see that your email shows as your username

**Expected behavior**
username should be validated in order to prevent an email to be added as a username

**Screenshots**
image

yaya-mamoudou:
## Summary
The user will no longer be able to input email as a username.

Closes #713 

## Changes

- Add a username validator in CustomRegisterSerializer


zubhub_backend/zubhub/creators/serializers.py
```diff
@@ -184,6 +184,15 @@ class CustomRegisterSerializer(RegisterSerializer):
     bio = serializers.CharField(allow_blank=True, default="", max_length=255)
     subscribe = serializers.BooleanField(default=False)
 
+    def validate_username(self, username):
+        # Check if the value is a valid email
+        pattern = r'^[\w\.-]+@[\w\.-]+\.\w+$'
+        if re.match(pattern, username):
+            # If it's a valid email, raise an error
+            raise serializers.ValidationError("Username cannot be an email.")
```
tuxology: Can we wrap string as `_("Username cannot be an email.")`. This would help in having the error translated in UI. Refer: https://docs.djangoproject.com/en/4.2/topics/i18n/translation/

tuxology (changes requested):

zubhub_backend/zubhub/creators/serializers.py
yaya-mamoudou: Hi @tuxology. This is done. Thanks for the share reference.

PR #672, opened 2023-03-15
Fixed scroll back to top button issue

ipriyanshi1708:
**This Pull request make the following changes:**
* Fixes issue #628  

**Checklist:**
- [x] I have created an issue for this solution I am submitting.
- [x] I created a new branch from the correct branch and made my changes in the new branch created.
- [x] My pull request is sent to the correct base branch in the original https://github.com/unstructuredstudio/zubhub repo.
- [x] I have linked the issue I worked on to this pull request submitted by me.
- [x] File(s) or Folder(s) now contain changes as specified in the issue I worked on.
- [x] I certify that I ran my checklist.

Ping @srish @tuxology @ajain1921 @NdibeRaymond 

Closes #628

PR #739, opened 2023-10-02
(fix) `Go-to-top` button shouldn't be visible if user is on top of a page

### LINKED ISSUE #628: Hide "Scroll back to top" button when already on top
Hi! @srish @tuxology, Scroll to top button should not be visible when user is already at the top of the website.

image

Thanks!

coderatomy:
## Summary

User shouldn't see the `go-to-top` icon/button once on top of the page
Closes #628

## Changes

- Added `useIsInViewPort` hook


coderatomy:
In continuation of the discussion in issue #672, and delving deeper into this topic, @tuxology, it appears that relying solely on a ref to a specific element might be insufficient, as pointed out by @chaitanya-cloudknox.

Consider this scenario: if we attach a `ref` to an element at the top of the home page, and utilizing the `useIsInViewPort` hook, the button displays correctly when the user scrolls down and the element is within the viewport. However, if the user navigates to another page, like the profile page, the `ref` element is out of the viewport. Consequently, the button won't appear as intended

zubhub_frontend/zubhub/src/hooks/useIsInViewPort.hook.js
```diff
@@ -0,0 +1,24 @@
+import { useState, useMemo, useEffect } from "react";
+
+
+export function useIsInViewport(ref) {
```
coderatomy: this hook generally watches over the reference element `ref` using the `IntersectionObserver` API. This will be used to check if an element is in the `view port`

tuxology:
@kamthamc Are you ok with this approach?

zubhub_frontend/zubhub/src/hooks/useIsInViewPort.hook.js
chaitanya-cloudknox: Can you move the IntersectionObserver creation to the useEffect as calling disconnect will stop emitting any new events?

Usually ref won't change, the value ref.current changes.  Just curious how the useEffect is working if the target element is remounted or re-rendered.

chaitanya-cloudknox:
> In continuation of the discussion in issue #672, and delving deeper into this topic, @tuxology, it appears that relying solely on a ref to a specific element might be insufficient, as pointed out by @chaitanya-cloudknox.
> 
> Consider this scenario: if we attach a `ref` to an element at the top of the home page, and utilizing the `useIsInViewPort` hook, the button displays correctly when the user scrolls down and the element is within the viewport. However, if the user navigates to another page, like the profile page, the `ref` element is out of the viewport. Consequently, the button won't appear as intended

I didn't understand it but the code looks fine. I have couple comments on it

zubhub_frontend/zubhub/src/hooks/useIsInViewPort.hook.js
tuxology: Any updates on this @coderatomy? This is a valid question

coderatomy: Sorry for the late reply @tuxology @chaitanya-cloudknox 

Ideally, the useEffect hook handles remounting or re-rendering of the target element by cleaning up the previous IntersectionObserver and creating a new one to observe the updated element, thanks to the dependency array [ref]. This ensures that the hook stays in sync with the current state of the DOM.

coderatomy:
My main concern here @chaitanya-cloudknox is that using `ref` won't be the best solution here, just because we have different pages.
To get this best @chaitanya-cloudknox, where do you think the ref should be attached?

chaitanya-cloudknox:
Tell me more about it.
I am assuming it is for the go to top button. 
My understanding is in the useEffect you create a listener with the ref.current dependency (since the current might change with re render and ref might not) and during unmount you disconnect the listeners.

If the user navigates to different page we won't be showing the go to top button until user scrolls to the bottom. If we check the case if current is not null/undefined and skip the part of adding event listeners would solve the problem. And also this hook will be only invoked if its consumed in the page, i am assuming we use this hook in each page separately

coderatomy:
> Tell me more about it. I am assuming it is for the go to top button. My understanding is in the useEffect you create a listener with the ref.current dependency (since the current might change with re render and ref might not) and during unmount you disconnect the listeners.
> 
> If the user navigates to different page we won't be showing the go to top button until user scrolls to the bottom. If we check the case if current is not null/undefined and skip the part of adding event listeners would solve the problem. And also this hook will be only invoked if its consumed in the page, i am assuming we use this hook in each page separately

I have changed to another approach. So basically, we wanted to show the button conditionally once the user is not at the top of the page. So adding the ref to the button itself is insufficient. I have updated to `window.pageYOffset`

Here I intend to show the button once there is any slight scroll down.
Simply, once a user is scrolling down, it is shown, otherwise not. I think this is easier and straight forward @chaitanya-cloudknox 

[Screencast from 2023-10-08 09-58-31.webm](https://github.com/unstructuredstudio/zubhub/assets/141822315/717554b1-1ba4-44f7-bf9e-5e4d7cf8164c)

tuxology (approved):

PR #744, opened 2023-10-03
(chore) Remove yarn.lock file

coderatomy:
## Summary

As per the conversation [HERE](https://unstructured.zulipchat.com/#narrow/stream/246561-ZubHub/topic/Technical.20support/near/394523327), it was confirmed that we should use `npm` as the package manager.
cc @srish 

## Changes

- Removed the `yarn.lock` file


tuxology (approved):
Thanks!

PR #745, opened 2023-10-03
Fixed input labels getting merged inside input borders when opened for the first time

### LINKED ISSUE #627: Email label is getting merged inside input border when open for the first time
**Describe the bug**
In the password reset section the email is getting merged inside input border
**To Reproduce**
Steps to reproduce the behaviour:
1. Go to [Password reset section](https://zubhub.unstructured.studio/password-reset)
2. Click on input 
3. Observe that email label getting merged inside border

**Screenshots**
Screenshot 2023-03-10 at 7 55 39 PM

brrkrmn:
## Summary

Input labels of login and password reset forms are merged into the input border when clicked for the first time. The problem is not seen in the sign up page because of the state changes that trigger rerender.

Closes #627 

## Changes

- Create `useForceUpdate` hook to trigger a render so that the correct width of the labels are accessible in login and password reset pages.

## Screenshots
Before:
Screenshot 2023-10-03 at 10 02 15

After:
Screenshot 2023-10-03 at 09 56 41


zubhub_frontend/zubhub/src/hooks/useForceUpdate.hook.js
```diff
@@ -0,0 +1,9 @@
+import { useEffect, useState } from 'react'
+
+export function useForceUpdate () {
```
tuxology: So, I think this type of fix will cause a lot of redraws when trying to attempt this fix. I don't feel that comfortable with this as of now. Is there any other creative way to fix this problem? How is it being handled in Google's form inputs?

tuxology (changes requested):
See comment below 👇

zubhub_frontend/zubhub/src/hooks/useForceUpdate.hook.js
brrkrmn: I am doing research on this, I will keep you posted 👍🏻

brrkrmn: Hey, as it turns out, the `labelWidth` prop that we were using was removed from MUI OutlinedInput. The new `label` prop calculates the width automatically. So I updated all OutlinedInputs to use the `label` prop.

Here is the link for this change:
https://mui.com/material-ui/migration/v5-component-changes/#outlinedinput

Can you review it again?

tuxology: Nice 🙌

tuxology (approved):
I think this should work - LGTM 🌈 Thanks!

PR #780, opened 2023-10-04
fix:add forgot password to `PageWrapper.jsx` so that it can be called when the user is not logged in

### LINKED ISSUE #741: Forget password option throws 404 error.
**Issue**
When I try to forget my password, it just throws a 404-error page. 

**To Reproduce**

1. Go to sign in page
2. Click on forget password option

**Expected behavior**
It should open page for forget password, so people can reset their password.

**Screenshots**
image

benndip:
## Summary

Fix the forgot flow by adding required routes to `PageWrapper.jsx`
Closes #741 

## Changes

- Include the following routes to the conditional statement for when the user is not logged in, and the route is not found
Screenshot 2023-10-04 at 20 10 22

- Include the following routes to the conditional statement for when the user is not logged in, and the route is found
Screenshot 2023-10-04 at 20 10 42

## Screenshots

Screenshot 2023-10-04 at 19 59 52

Screenshot 2023-10-04 at 20 00 07

### DIFF
--- a/zubhub_frontend/zubhub/src/views/PageWrapper.jsx
+++ b/zubhub_frontend/zubhub/src/views/PageWrapper.jsx
@@ -236,6 +236,9 @@ function PageWrapper(props) {
             '/terms_of_use',
             '/about',
             '/challenge',
+            '/password-reset',
+            '/email-confirm',
+            '/password-reset-confirm'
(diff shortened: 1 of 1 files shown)


tuxology (approved):
We should take this list from elsewhere eventually, for now fine here 👍

PR #787, opened 2023-10-04
feat: implements cursor display on hover for notif

### LINKED ISSUE #782: Cursor not turning into pointer on hover on notification and profile icons
**Describe the bug**
In the Navbar section, when I hover over the icons for Notification and Profile, it does not change the cursor into a pointer. For new users, this can be confusing and they might not how to access the profile section. 
This becomes especially confusing because there is a part on the navbar that shows the username. This part is not clickable and so the cursor being default is right. 
A new user might get confused and can keep clicking on the username to access the profile section. Only after some trials can they click on the profile icon and navigate to the profile section. This is a small issue but disrupting the UX

**To Reproduce**
Go to Zubhub page and hover over the profile and notification icons on the Navbar

**Expected behavior**
When we hover over the language icon or the search bar the cursor changes into a pointer indicating that some action can be performed. This is expected of the profile and notification icon too.

**Screenshots**
Screenshot from 2023-10-05 01-26-16
In the screenshot, the cursor is not visible but we can try hovering on all the clickable icons and see the bug.

yokwejuste:
## Summary

Description of PR here...
Closes #782

## Changes
- https://github.com/yokwejuste/zubhub/77146fb2a71737cea9e2bbef9c57c36b5de48b3d/zubhub_frontend/zubhub/src/components/notification_button/NotificationButton.jsx?plain=1#L21

## Screenshots
Here is the output:

https://github.com/unstructuredstudio/zubhub/assets/71908316/f0faa599-b646-4edd-99b9-7005d2eca764

### DIFF
--- a/zubhub_frontend/zubhub/src/components/avatar_button/AvatarButton.jsx
+++ b/zubhub_frontend/zubhub/src/components/avatar_button/AvatarButton.jsx
@@ -41,6 +41,7 @@ const AvatarButton = props => {
           className={commonClasses.iconBox}
           alt={auth?.username?.toUpperCase()}
           src="/static/images/avatar/1.jpg"
+          style={{ cursor: 'pointer' }}
         />
         <PanelPopper open={dropdownOpen} anchorEl={buttonRef} signUpStyles={signupStyles}>
           <div className={classes.container}>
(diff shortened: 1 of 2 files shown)


aqsaaqeel:
This only fixes the cursor for notification icon but we need for avatar icon too

yokwejuste:
The left out Avatar cursor is now enabled @srish @tuxology kindly review this

tuxology (approved):
Thanks @yokwejuste 🌈

yokwejuste:
> Thanks @yokwejuste 🌈 

My pleasure 😌

PR #819, opened 2023-10-06
fix: removes tags selection duplicate

### LINKED ISSUE #776: Improve the matching algorithm for materials field
image

When typing and hitting enter, it adds the typed material but the input value stays same and not cleared.

yokwejuste:
## Summary

Description of PR here...
Closes #776 

## Changes

The tag input file was modified:
```jsx
// On press of the Enter key
  const handleKeyDown = (e) => {
    if (e.key === 'Enter') {
      const inputValue = ref.current.value.trim();
      const [errMsg, isValid] = validateTag(inputValue, selectedTags);
      setErrorMessage(errMsg);
      if (isValid) {
        addTag(inputValue);
        setIsSearching(false);
        setErrorMessage('');
        ref.current.value = '';
      }
    }
  }; 
```
```jsx
// Little modif for the click away function
  const handleClickAway = () => {
    const inputValue = ref.current.value.trim();
    if (!inputValue || selectedTags.includes(inputValue)) {
      setIsSearching(false);
      setErrorMessage('');
    } else {
      const isFound = remoteData.some(tag => tag.name === inputValue);
      if (isFound || !remoteData.length) {
        addTag(inputValue);
        setIsSearching(false);
        setErrorMessage('');
        ref.current.value = '';
      } else {
        setErrorMessage(`The tag "${inputValue}" was not found and cannot be added.`);
      }
    }
  };
```
```jsx
// Error message
{errorMessage && <p style={{ color: 'red' }}>{errorMessage}</p>}
```
```jsx

// Tag validator
  const validateTag = (tag, selectedTags) => {
    const trimmedTag = tag.trim();
    if (trimmedTag === '') {
      return ['Tag cannot be empty.', false];
    } else if (selectedTags.includes(trimmedTag)) {
      return [`The tag "${trimmedTag}" is already in the search bar and cannot be added again.`, false];
    } else {
      return ['', true];
    }
  };

```
## Screenshots
Recording 2023-10-06 104112
image


yokwejuste:
@srish @tuxology can you have a look at this ?

zubhub_frontend/zubhub/src/components/form/tagsInput/TagsInput.jsx
```diff
@@ -70,20 +112,20 @@ export default function TagsInput({
     </CustomButton>
   ));
 
-  selectedTags = selectedTags?.map((tag, index) => (
+  const selectedTagsRender = selectedTags?.map((tag, index) => (
```
coderatomy: consider returning a memorized version of this

zubhub_frontend/zubhub/src/components/form/tagsInput/TagsInput.jsx
```diff
@@ -70,20 +112,20 @@ export default function TagsInput({
...
       endIcon={<ClearRounded />}
     >
-      {prefix && `${prefix}`}
+      {prefix && `${prefix} `}
       {tag}
     </CustomButton>
   ));
 
-  remoteData = remoteData?.map((tag, index) => (
+  const remoteDataRender = remoteData?.map((tag, index) => (
```
coderatomy: same as above

coderatomy (changes requested):
Thanks @yokwejuste for this. Just a few comments

zubhub_frontend/zubhub/src/components/form/tagsInput/TagsInput.jsx
yokwejuste: Hi, 

Please can you give me more highlights of this?

zubhub_frontend/zubhub/src/components/form/tagsInput/TagsInput.jsx
yokwejuste: I'm a little failing to understand you on this, please elaborate

zubhub_frontend/zubhub/src/components/form/tagsInput/TagsInput.jsx
coderatomy:
```
const selectedTagsRender = useMemo(() => {
    return selectedTags?.map((tag, index) => {
        // JSX rendering logic here
    });
}, [selectedTags, /* dependencies */]);
```

It's ideally a best practice `perfomance wise` to prevent unnecessary re-renders 
Checkout [here](https://www.linkedin.com/pulse/avoid-unnecessary-re-renders-reactjs-christian-zanchetta) for more

yokwejuste:
Hey @coderatomy I did some updates on the code:
After the point you made, it called my attention on the whole component I was working on, and the updates below were made:
## Updates and Review

### Optimization
- **useMemo**: I used useMemo to memorize the rendering of some components (the tags and data lists). Normally, these components would be re-rendered every time our component updates, even if the tags haven't changed. useMemo "remembers" the rendered components and only re-renders them if the tags/data they depend on change.

- **useCallback**: I used useCallback to memorize functions, so they keep their identity across re-renders unless their dependencies change. Normally, functions are recreated every time the component re-renders, and if these functions are used in effects or passed to children, it may cause unnecessary re-renders or effect runs.

### Impact on the Code and Performance:
- Improved Performance: The primary impact is improved performance, especially as the list of tags or data grows. By not re-rendering components or recalculating values that haven’t changed, we save processing time and make our app smoother for the user.

- Reduced Unnecessary Updates: Using memoized functions and components means fewer updates to the virtual and actual DOM, making the rendering process more efficient and improving the user experience, particularly in larger applications or on slower devices.

- Cleaner Code (to an extent): The logic for rendering and data management is separated, and clearer, making it slightly easier to understand the rendering logic separately from the data manipulation or action logic.

@srish @tuxology I'll like you to have a look over this PR modifcations.

tuxology:
It is hard to review this code @yokwejuste Can you create two clean commits - one with linting changes and one with actual changes so I can separate one from another?

yokwejuste:
> It is hard to review this code @yokwejuste Can you create two clean commits - one with linting changes and one with actual changes so I can separate one from another?

sure :)

yokwejuste:
@tuxology the linting was removed and only the necessary changes were left out.

tuxology:
@kamthamc Can you review this?

zubhub_frontend/zubhub/src/components/form/tagsInput/TagsInput.jsx
```diff
@@ -28,66 +28,107 @@ export default function TagsInput({
...
-  const handleChange = e => {
-    setIsSearching(true);
-    onChange(e.target.value);
-  };
+  const validateTag = useCallback((tag, selectedTags) => {
+    const trimmedTag = tag.trim();
+    if (trimmedTag === '') {
+      return ['Tag cannot be empty.', false];
+    } else if (selectedTags.includes(trimmedTag)) {
+      return [`The tag "${trimmedTag}" is already in the search bar and cannot be added again.`, false];
```
chaitanya-cloudknox: this needs translation, And also it doesn't depend on any props, so move it out of the component

zubhub_frontend/zubhub/src/components/form/tagsInput/TagsInput.jsx
```diff
@@ -28,66 +28,107 @@ export default function TagsInput({
...
+    }
+  }, []);
 
-  const handleClickAway = () => setIsSearching(false);
-  const handleFocus = () => {
-    clearSuggestions();
-  };
+  const handleTagAddition = useCallback((value) => {
+    const [errMsg, isValid] = validateTag(value, selectedTags);
+    setErrorMessage(errMsg);
```
chaitanya-cloudknox, line 50: Move it to else block. Even though it doesn't rerender immediately, its good to avoid setting same states immediately

zubhub_frontend/zubhub/src/components/form/tagsInput/TagsInput.jsx
```diff
@@ -28,66 +28,107 @@ export default function TagsInput({
...
+      setErrorMessage('');
+    } else {
+      const isFound = remoteData.some(tag => tag.name === inputValue);
+      if (isFound || !remoteData.length) {
+        addTag(inputValue);
+        setIsSearching(false);
+        setErrorMessage('');
+        ref.current.value = '';
+      } else {
+        setErrorMessage(`The tag "${inputValue}" was not found and cannot be added.`);
```
chaitanya-cloudknox: missing translation

zubhub_frontend/zubhub/src/components/form/tagsInput/TagsInput.jsx
```diff
@@ -28,66 +28,107 @@ export default function TagsInput({
...
 
-  popularTags = popularTags?.map((tag, index) => (
+  const popularTagsRender = useMemo(() => popularTags?.map((tag, index) => (
     <CustomButton
       onClick={() => handleTagAddition(tag)}
       disabled={selectedTags.includes(tag)}
       className={clsx(classes.button, selectedTags.includes(tag) && classes.disabledButton)}
-      style={{ fontWeight: tag == 'General' && '800' }}
-      primaryButtonOutlinedStyle
+      style={{ fontWeight: tag === 'General' ? '800' : 'normal' }}
```
chaitanya-cloudknox: use classes instead of inline styles

zubhub_frontend/zubhub/src/components/form/tagsInput/TagsInput.jsx
```diff
@@ -28,66 +28,107 @@ export default function TagsInput({
...
       startIcon={<Add />}
     >
       {prefix && `${prefix} `}
       {tag}
     </CustomButton>
-  ));
+  )), [popularTags, handleTagAddition, selectedTags, prefix, classes]);
 
-  selectedTags = selectedTags?.map((tag, index) => (
+  const selectedTagsRender = useMemo(() => selectedTags?.map((tag, index) => (
```
chaitanya-cloudknox: Memos not required, are they causing any performance issues? How large can the size of the array be?

same applies for the one below

zubhub_frontend/zubhub/src/components/form/tagsInput/TagsInput.jsx
```diff
@@ -115,14 +156,15 @@ export default function TagsInput({
         <Box style={{ position: 'relative' }}>
           <Box className={classes.suggestionBox}>
             <ClickAwayListener onClickAway={handleClickAway}>
-              <Box>{remoteData.length > 0 ? remoteData : <p>Item not found: Hit Enter to Save your Input</p>}</Box>
+              <Box>{remoteData.length > 0 ? remoteDataRender : <p>Item not found: Hit Enter to Save your Input</p>}</Box>
```
chaitanya-cloudknox: Missing translation

zubhub_frontend/zubhub/src/components/form/tagsInput/TagsInput.jsx
```diff
@@ -115,14 +156,15 @@ export default function TagsInput({
...
             </ClickAwayListener>
           </Box>
         </Box>
       ) : null}
 
+      {errorMessage && <p style={{ color: 'red' }}>{errorMessage}</p>}
```
chaitanya-cloudknox: use a common class for errors

zubhub_frontend/zubhub/src/views/create_project/step1/Step1.jsx
```diff
@@ -102,7 +102,7 @@ function Step1({ formik, handleBlur }) {
           clearSuggestions={clearSuggestions}
           removeTag={removeTag}
           handleBlur={formik.handleBlur}
-          placeholder="Start typing to materials used"
+          placeholder="Start typing your materials used"
```
chaitanya-cloudknox, line 105: this needs translation

yokwejuste:
@chaitanya-cloudknox thanks for the review, will look into that

zubhub_frontend/zubhub/src/components/form/tagsInput/TagsInput.jsx
tuxology: See existing ways in which errors are displayed @yokwejuste

zubhub_frontend/zubhub/src/components/form/tagsInput/TagsInput.jsx
yokwejuste: As for this the translations where not part of this issue, so to be moved to the next PR :)

zubhub_frontend/zubhub/src/components/form/tagsInput/TagsInput.jsx
yokwejuste: Even if the tags array isn't significantly large, the computation or JSX rendering might be complex. Using useMemo ensures that these computations are only done, when necessary, thus potentially improving performance. and for Futureproofing will be good to start using this now. But your suggestion makes a lot of sense, but do we agree this makes it a little more optimized?

zubhub_frontend/zubhub/src/views/create_project/step1/Step1.jsx
yokwejuste: As previously said, translation for this case is not with respect to this issue. @tuxology will be advanceable to do the translation on this PR, I'm trying to keep it as focused as possible.

zubhub_frontend/zubhub/src/components/form/tagsInput/TagsInput.jsx
kamthamc: Do not use useMemo unless its affecting performance. its gonna increase memory and cause of bugs, We could ignore minor performance gains.

> Premature Optimization Is the Root of All Evil

yokwejuste:
Hi @tuxology @chaitanya-cloudknox @coderatomy done with the corrections here.

The updates are as follows:
- Remove useMemo after research since we are dealing with a small data size.
- Applying translations
- Moving from inline styling to classes

zubhub_frontend/zubhub/src/components/form/tagsInput/TagsInput.jsx
yokwejuste-surs: all god here

zubhub_frontend/zubhub/src/components/form/tagsInput/TagsInput.jsx
yokwejuste-surs: the styles were migrated to classes

zubhub_frontend/zubhub/src/components/form/tagsInput/TagsInput.jsx
yokwejuste: > Do not use useMemo unless its affecting performance. its gonna increase memory and cause of bugs, We could ignore minor performance gains.
> 
> > Premature Optimization Is the Root of All Evil

Got it now!

I did some changes, please help me review the code

tuxology (approved):
LGTM 🌈

PR #825, opened 2023-10-06
(feat) Implement a reusable route protection component

coderatomy:
## Summary
Instead showing of show a `404` error page to routes that do exist but actually just require authentication ie for a user to be logged in.
Closes 

## Changes

- Added `route protection` component

## API
- **Path**: `string`
     Route to which the a given component will be rendered

- **Component**: `React.component`
     A react component that's to be rendered with the `path`

`<ProtectedRoute path='/home' component={Home} {...props} />`

## Notes
  - Don't forget to path a distructured `props` object. It contains vitals properties for the application to run

## Screencasts
[Screencast from 2023-10-06 16-05-59.webm](https://github.com/unstructuredstudio/zubhub/assets/141822315/4147f8c1-3f30-417d-a004-ef5cb50b4f5c)

### DIFF
--- a/zubhub_frontend/zubhub/src/App.js
+++ b/zubhub_frontend/zubhub/src/App.js
@@ -1,97 +1,50 @@
-import React, { useContext }  from 'react';
+import React, { useContext } from 'react';
 import { withTranslation } from 'react-i18next';
-import { BrowserRouter as Router, Route, Switch } from 'react-router-dom';
-import CreateActivity from './views/create_activity/CreateActivity';
-import { useEffect, useState } from "react";
+import { BrowserRouter as Router, Route, Redirect, Switch } from 'react-router-dom';
+import CreateActivity from './views/create_activity/create_activity';
+import { useEffect } from 'react';
+import { connect } from 'react-redux';
 import LoadingPage from './views/loading/LoadingPage';
 import PageWrapper from './views/PageWrapper';
 import ZubhubAPI from '../src/api/api';
 import { updateTheme } from './theme';
 import ScrollToTop from './ScrollToTop';
 
-const SearchResults = React.lazy(() =>
-  import('./views/search_results/SearchResults'),
-);
+const SearchResults = React.lazy(() => import('./views/search_results/SearchResults'));
 
 const Signup = React.lazy(() => import('./views/signup/Signup'));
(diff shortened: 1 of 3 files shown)


coderatomy:
For now, I have implemented this to only the `/profile`, `/settings` and `/team` which I was kinda sure needed authentication. Am to continue adding this to more routes. Am just curious to know from folks like @srish @tuxology @iamhks which routes are sure need this feature.

zubhub_frontend/zubhub/src/App.js
```diff
@@ -112,10 +64,26 @@ const LazyImport = props => {
   );
 };
 
+const ProtectedRoute = ({ component: LazyComponent, ...props }) => {
```
coderatomy: This checks if the user is authenticated by accessing the `token` and redirects to the login page if not.

zubhub_frontend/zubhub/src/App.js
```diff
@@ -175,555 +134,401 @@ function App(props) {
...
+          <Route
+            path="/teams/:groupname"
+            render={routeProps => (
+              <PageWrapper {...routeProps} {...props}>
+                <LazyImport LazyComponent={Team} {...routeProps} {...props} />
+              </PageWrapper>
+            )}
+          />
+
+          <ProtectedRoute path="/profile" component={Profile} {...props} />
```
coderatomy, line 333: using the `ProtectRoute` component. This is where I made all the changes. The rest are prettier indention updates. Am also hoping to configure `prettier` with `eslint` for code uniformity
cc @tuxology

tuxology:
@coderatomy Thanks for implementing this initial version. However, there is regarding UX approaches that can be taken before we begin implementing this. It is possible that this component will be used, but equally possible that we won't use this. Let's finish the discussion and finalize the UX before working on this 👍 **Each PR should be linked to an issue where a consensus is achieved**

coderatomy:
> @coderatomy Thanks for implementing this initial version. However, there is regarding UX approaches that can be taken before we begin implementing this. It is possible that this component will be used, but equally possible that we won't use this. Let's finish the discussion and finalize the UX before working on this 👍 **Each PR should be linked to an issue where a consensus is achieved**

No issues. Thanks for informing 🙂

tuxology (approved):

PR #833, opened 2023-10-07
feat[UI/Functionality]: Add navigation buttons to the opened dialog carousel images

benndip:
## Summary

Add navigation buttons to the opened dialog carousel images so that users can still scroll through to view the images without needing to close the dialog first(improved UX)
Closes #435

## Changes

- Followed the UI/UX design and logic made on Figma by (Steph Uzoukwu)  [https://github.com/Stephxdesigns](url)
- Built the components and added the images from the Figma design
- Write the functionality in `projectDetailScripts.js` to change the images when the Navigation buttons a clicked in 

## Demo

https://github.com/unstructuredstudio/zubhub/assets/51094040/48321e3b-8639-496f-a7e2-31437390e304

PR #842, opened 2023-10-08
Improved navigation button on carousel images

### LINKED ISSUE #435: Navigation buttons on the caraousel images.

brrkrmn:
## Summary

The navigation buttons on the carousel images are improved, based on the changes in #613 and @kamthamc's reviews. 

Closes #435

## Changes

- Classes are used instead of inline styles, only with the exception of `NextArrow` and `PrevError` function in `projectDetailsStyles.js`. The current way is also used in the React Slick's documentation. Using classes instead of inline styles overrides, so I did not changed that part.
- Aria label for the carousel close button is added to the translation files.
- Theme colors are used where these changes were made - the rest of the file still involves hardcoded colors -

## Final version

https://github.com/unstructuredstudio/zubhub/assets/92817363/0e5919e3-2bb1-4f35-bf35-8e3f7a422ff4


zubhub_frontend/zubhub/src/views/project_details/ProjectDetails.jsx
```diff
@@ -428,11 +431,15 @@ function ProjectDetails(props) {
 
         <Dialog
           PaperProps={{
-            style: {
-              backgroundColor: 'transparent',
-              boxShadow: 'none',
-            },
+            className: classes.enlargedImageDialogPaperStyle,
           }}
+          BackdropProps={{
```
tuxology: Why do we need this? If its not used, we can remove this

tuxology:
@kamthamc Requesting you to look at #842 and #833  We wish to pick and optimal implementation from these

zubhub_frontend/zubhub/src/assets/js/styles/views/project_details/projectDetailsStyles.js
```diff
@@ -189,15 +190,48 @@ const styles = theme => ({
...
-    maxWidth: '600px',
+    padding:'2px',
+    width: '85%',
+    maxWidth: '840px',    
     height: 'auto',
   },
+  cancelEnlargedImageBtn: {
+    backgroundColor: colors.secondary,
+    position: 'absolute',
+    right: '60px',
```
kamthamc: I want to check how it looks on mobile, I am guessing its too far from right corner

zubhub_frontend/zubhub/src/assets/js/styles/views/project_details/projectDetailsStyles.js
```diff
@@ -400,4 +434,28 @@ export function PrevArrow(props) {
   );
 }
 
+export function EnlargedImgArrow (props) {
+  const { className, style, onClick } = props;
+  return (
+    <div
+      className={className}
+      style={{
```
kamthamc: avoid inline styles, instead use use the classes. if some style needs to be set dynamically use css properties

zubhub_frontend/zubhub/src/views/project_details/ProjectDetails.jsx
```diff
@@ -324,14 +327,14 @@ function ProjectDetails(props) {
...
-                        {project.images.map(image => (
+                        {project.images.map((image, index) => (
                           <div>
                             <img
                               key={image.public_id}
                               className={classes.carouselImageStyle}
                               src={image.image_url}
                               alt={image.public_id}
-                              onClick={e => handleSetState(handleOpenEnlargedImageDialog(e, state))}
+                              onClick={e => handleSetState(handleOpenEnlargedImageDialog(e, state, index))}
```
kamthamc, line 337: Its better to use either the image ref or image id rather than index. In cases like slicing array or infinite scrolling it might affect indices.

zubhub_frontend/zubhub/src/views/project_details/ProjectDetails.jsx
```diff
@@ -443,7 +450,52 @@ function ProjectDetails(props) {
           }
           aria-labelledby={t('projectDetails.ariaLabels.imageDialog')}
         >
-          <img className={classes.enlargedImageStyle} src={enlarged_image_url} alt={`${project.title}`} />
+          <IconButton
+            className={classes.cancelEnlargedImageBtn}
+            onClick={() =>
+              setState({
```
kamthamc: use setState function instead
```javascript
setState(oldState => ({...oldState, newProp: 'value'}));
```

zubhub_frontend/zubhub/src/views/project_details/ProjectDetails.jsx
```diff
@@ -443,7 +450,52 @@ function ProjectDetails(props) {
...
+                <Slider
+                  initialSlide={enlarged_image_id}
+                  adaptiveHeight
+                  infinite
+                  speed={500}
+                  slidesToShow={1}
+                  slidesToScroll={1}
+                  className={classes.enlargedImageStyle}
+                  nextArrow={<EnlargedImgArrow />}
+                  prevArrow={<EnlargedImgArrow />}
```
kamthamc, line 476: I don't know how the arrow looks, I am assuming they look like `<`, `>`. I think direction of arrow is missing

zubhub_frontend/zubhub/src/views/project_details/ProjectDetails.jsx
```diff
@@ -443,7 +450,52 @@ function ProjectDetails(props) {
...
+                >
+                  {project.images.map(image => (
+                    <div>
+                      <img
+                        className={classes.sliderImageStyle}
+                        key={image.public_id}
+                        src={image.image_url}
+                        alt={''}
```
kamthamc: @tuxology I think we should ask optional image descriptions while uploading images for accessibility

kamthamc (commented):
I have some minor comments, I looks good so far

zubhub_frontend/zubhub/src/assets/js/styles/views/project_details/projectDetailsStyles.js
brrkrmn: @kamthamc Yes, it is too far. I can make it `right: '6%'` which works well for all sizes, or I can use media queries. Which one should I go with?

chaitanya-cloudknox: You can use something like clamp 

https://developer.mozilla.org/en-US/docs/Web/CSS/clamp

zubhub_frontend/zubhub/src/views/project_details/ProjectDetails.jsx
brrkrmn: I tried to use id, however, react slick's initialSlide property requires the index of the image as an int, and our images have id's that look like this: `project_images/qgAS64S-ZkYwqL5RKve7d`. Thus, it does not work. I searched for other ways but using the index for this prop is the only way I found. The index is currently only used for the initialSlide prop.

zubhub_frontend/zubhub/src/assets/js/styles/views/project_details/projectDetailsStyles.js
brrkrmn: I am afraid I do not understand why we should not go with 6%. It makes the button dynamically placed at the top right corner for every screen size. Should we not use relative units like percentages that are set relative to some other value? Since we do not have a preferred value here, I don't think using clamp with 3 specific values is the right solution.

zubhub_frontend/zubhub/src/views/project_details/ProjectDetails.jsx
brrkrmn: React slick deals with the arrows themselves, the only styling I did is the background color, border radius of the circles etc. Using nextArrow, and prevArrow props automatically makes the arrows face the right directions.

zubhub_frontend/zubhub/src/assets/js/styles/views/project_details/projectDetailsStyles.js
kamthamc: For example if you open an application in your machine, the close button won't be responsive (the position won't change) that has to be placed at the top right with some gaps (margins). There are some elements that can move based on responsiveness and some elements should not move/adjust based on screen sizes.

zubhub_frontend/zubhub/src/views/project_details/ProjectDetails.jsx
tuxology: I agree, accessibility work will happen later as a completely new app-wide project - so out of scope for now

zubhub_frontend/zubhub/src/views/project_details/ProjectDetails.jsx
tuxology: @brrkrmn Can you resolve this, rest LGTM

brrkrmn:
@kamthamc sorry it took so long, can you review the final version?

zubhub_frontend/zubhub/src/assets/js/styles/views/project_details/projectDetailsStyles.js
```diff
@@ -347,57 +389,81 @@ export const sliderSettings = images_num => ({
...
+      style={nextArrowStyle}
       onClick={onClick}
     />
   );
 }
 
 export function PrevArrow(props) {
-  const { className, style, onClick } = props;
+  const { className, onClick } = props;
+  const prevArrowStyle = {
```
chaitanya-cloudknox: I mean to move it to makeStyles, so it creates classnames

brrkrmn: When I added these next to other styles in this file, I could not use the className to style the component. I tried giving the className in projectDetails.jsx and projectDetailsStyles.js as well but could not achieve the expected result. Should I import makeStyles in this file to create the classnames?

brrkrmn: updated this part using makeStyles

tuxology (approved):

PR #843, opened 2023-10-08
Fix: add translation keys for hardcoded texts

### LINKED ISSUE #222: Auto translation of whole ZubHub in multiple languages
We need this feature to support translation of static as well as dynamic data .
* Static (e.g., header on home page, nav bar titles, description in about page, etc.)
* Dynamic (e.g., projects, comments, user profile, etc.)

Perhaps something of this sort like Facebook does:
Screen Shot 2021-07-03 at 11 30 46 AM.

Currently, we are handling this well for some elements (buttons, footer menu items). Ideally, the app should support for translating content for every single element in multiple languages.

brrkrmn:
## Summary

Some parts of the code were using hardcoded text. This PR adds translation keys for those texts.

Closes #222


srish:
Related to https://github.com/unstructuredstudio/zubhub/issues/222

tuxology (approved):
We can merge this 🤔 There might be other locations as well.

PR #844, opened 2023-10-08
fix: match overlay border radius to card's

brrkrmn:
## Summary

The overlay border had a mismatch border radius. This PR addresses that.

## Changes

- Changes overlay border radius to 15px.

## Videos

Before:

https://github.com/unstructuredstudio/zubhub/assets/92817363/a433e034-2af5-4104-bba6-aefc8253a30d

After:

https://github.com/unstructuredstudio/zubhub/assets/92817363/a7c56f6d-ca03-487c-919d-82ca02741877


zubhub_frontend/zubhub/src/views/create_team/step3/step3.styles.js
```diff
@@ -23,7 +23,7 @@ export const step3Style = theme => ({
     },
     overLay:{height:'100%', width:'100%',top:0, position:'absolute', backgroundColor:'transparent', zIndex:1},
     blurOverLay:{
-        borderRadius: '6.93px',
+        borderRadius: '15px',
```
coderatomy, line 26: Can you please attach UI differences `before  and after` the changes in the summary above. It's easier for reviewers. 

We actually have to run this locally and test these changes to confirm if it is actually what you say it is. So attaching some screenshots is crucial more so for such UI changes

coderatomy (changes requested):
Nice work @brrkrmn.

coderatomy (approved):

tuxology (approved):

PR #845, opened 2023-10-08
feat: project error message

brrkrmn:
## Summary

Disabling the submit button is generally considered an antipattern because the user doesn't get any feedback about why the button is disabled. That is why I enabled the button by default and if no project type is selected, we show an error text.

## Screenshots
Before:

image

After:

image


zubhub_frontend/zubhub/src/views/create_project/CreateProject.jsx
```diff
@@ -430,11 +434,11 @@ const SelectModeUI = ({ setMode }) => {
             <Typography>If you worked on the project with other creators </Typography>
           </div>
         </div>
+        {isError && <div className={classes.projectSelectError}>You need to select project type.</div>}
         <CustomButton
           onClick={handleCreate}
           primaryButtonStyle
           style={{ marginTop: 40, alignSelf: 'center' }}
-          disabled={mode.length === 0}
```
coderatomy, line 437: Can you keep the button disabled? Some big-headed users might decide to click Create project when they haven't selected a mode. We generally want to limit them to that point until they choose one

coderatomy (approved):
Nice work @brrkrmn. Just a few suggestions and this is good to go

zubhub_frontend/zubhub/src/views/create_project/CreateProject.jsx
brrkrmn: Hey, with these changes, it is still not possible to create a project without selecting a mode. 
Keeping the error while disabling the button may not be possible because we show the error when the user clicks on the button, when it is disabled, the error would not be shown as well. So I guess we have to choose either

coderatomy: Alright. I think its fine then

coderatomy (approved):

srish:
@brrkrmn Hi! Could you please file an issue around this PR on GitHub and get a designer's input on the design flow? Good catch and TY! :)

brrkrmn:
> @brrkrmn Hi! Could you please file an issue around this PR on GitHub and get a designer's input on the design flow? Good catch and TY! :)

Sure, thank you

brrkrmn:
closing this since the error message will not be implemented. opened #894 instead

PR #849, opened 2023-10-09
fix sidenav to view port

aqsaaqeel:
## Summary

Fixed Sidenav to the viewport. User does not have to scroll all the way up to create a project. 

## Changes

- Sidenav position fixed.
- Sidenav height reduced so it does not collapse with footer.

## Screenshots

Screenshot from 2023-10-09 17-04-04


tuxology:
@aqsaaqeel We would like to consider this change later since the UX for sidebar may change when we make holistic changes. Closing this for now.

tuxology:
Hey @aqsaaqeel we are now considering this now. @srish and I were not on same page regarding this hence I closed this one!

tuxology (approved):

PR #855, opened 2023-10-10
fix[functionality]: Enable non logged in users to search for projects.

### LINKED ISSUE #802: Search is broken for non-logged in users
**Issues*

**To Reproduce**

**Expected behavior**

**Screenshots**

benndip:
## Summary

Description of PR here...
Closes #802

## Changes

- Edited `Navbar.jsx` to check whether `props.auth` exists and if not, we return all the `completions` 

## Demo

https://github.com/unstructuredstudio/zubhub/assets/51094040/27fb0354-0094-4194-a372-9cf84b871909

### DIFF
--- a/zubhub_frontend/zubhub/src/components/Navbar/Navbar.jsx
+++ b/zubhub_frontend/zubhub/src/components/Navbar/Navbar.jsx
@@ -113,7 +113,7 @@ function PageWrapper(props) {
         } else if (searchType === SearchType.PROJECTS) {
           completions = await api.autocompleteProjects({ query, token });
           completions = completions
-          .filter(c=>( c.creator.id === props.auth.id && c.publish.type !== 1 ))
+          .filter(c=> props.auth.id ? (c.creator.id === props.auth.id && c.publish.type !== 1): c )
           .map(({ id, title, creator, images }) => ({
             title,
             shortInfo: creator.username,
--- a/zubhub_frontend/zubhub/src/components/Navbar/Navbar.jsx
+++ b/zubhub_frontend/zubhub/src/components/Navbar/Navbar.jsx
@@ -113,7 +113,7 @@ function PageWrapper(props) {
         } else if (searchType === SearchType.PROJECTS) {
           completions = await api.autocompleteProjects({ query, token });
           completions = completions
-          .filter(c=> props.auth.id ? (c.creator.id === props.auth.id && c.publish.type !== 1): c )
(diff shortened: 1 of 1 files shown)


zubhub_frontend/zubhub/src/components/Navbar/Navbar.jsx
```diff
@@ -77,6 +77,8 @@ function PageWrapper(props) {
   const formRef = useRef();
   const token = useSelector(state => state.auth.token);
 
+  console.log("TOken", token);
```
brrkrmn: Good job, you forgot to delete this 🙂

benndip: oooooh, that's true. Thanks for noting that @brrkrmn , added that when trying to debug when the projects weren't showing. Let me remove it.

zubhub_frontend/zubhub/src/components/Navbar/Navbar.jsx
```diff
@@ -113,7 +113,7 @@ function PageWrapper(props) {
         } else if (searchType === SearchType.PROJECTS) {
           completions = await api.autocompleteProjects({ query, token });
           completions = completions
-          .filter(c=>( c.creator.id === props.auth.id && c.publish.type !== 1 ))
+          .filter(c=> props.auth.id ? (c.creator.id === props.auth.id && c.publish.type !== 1): c )
```
coderatomy: I believe it is `props.auth.token` that's what we refer to for authentication. I believe the `id` property is not in the auth object

coderatomy (changes requested):
Thanks for your effort. Just a few fix up. Am just curious if this was approved by @srish

zubhub_frontend/zubhub/src/components/Navbar/Navbar.jsx
benndip: Thanks for that remark @coderatomy . I have adjusted that to `props.auth.token` even though there is actually an `id` property in the auth object, I verified that. Also, @srish marked this as `good-first` and `high-priority` so that's why I selected it to work on it.

tuxology:
@benndip Thanks for implementing this initial version. However, there is regarding UX approaches that can be taken before we begin implementing this. There are three propositions we have in the description of #802. Until those are decided upon with help of UX and consensus is reached, we cannot proceed. Closing this.

PR #778, opened 2023-10-04
Feature add/edit team members

Formasitchijoh:
### Summary

This pull request addresses the feature of adding, deleting, and updating members within a specific team in the Zubhubs project. The addition of this functionality allows for better management of team members, enabling the ability to add new members, remove inactive ones, and update member roles.

###  Context
Issue #734 highlighted the need to implement the functionality of adding, deleting, and updating team members in the Zubhubs project. This feature is essential as it allows for the addition of members to a team after its creation. Additionally, it provides the capability to remove members who are no longer active or wish to be removed. Furthermore, verified admins can update member roles, transitioning them between member and admin positions. This pull request aims to implement this feature and enhance the overall management of team members.

### Details

Most of the required APIs for this feature were already in place, but some modifications were necessary to resolve encountered errors. A new folder called "add_team_member" was created, which contains the following files: "AddTeamMember.jsx" for updating the UI, "addTeamMemberScript" for handling HTTP requests and actions from the UserAction in the Redux store, and an updated "userFollower.css" file. Currently, efforts are focused on improving the user experience of this new feature and writing test cases.

### Changes

1.     Updated the "TeamMembers.jsx" file and improved the styling of the cards.
2.   Created a new folder called "add_team_member" in the "view" folder, which includes the "AddTeamMembers.jsx" file and "addTeamMembersScript" for implementing the add, edit, and delete functionality.

## Screenshots

Below is a video showcasing the implemented functionality.

https://github.com/unstructuredstudio/zubhub/assets/76694292/59be58f6-5b9f-4d8c-b7a4-6100c024bf29

@iamhks  @tuxology  @srish 
please kindly review this pull request when you  are chanced.

PR #857, opened 2023-10-10
(feat) Improve EditTeam flow 

brrkrmn:
## Summary

The EditTeam view's code involved unused functions and variables, and the design was not consistent with the EditProfile view. Additionally, I implemented the code to show error message when the user tries to submit the form without a  team name. 

## Changes for the appearance

- Edit Team button label is changed from "Edit Team Info" to "Edit Team"
- The input headings were removed since the labels convey the same text
- A description text is added for the bio field
- Validation is added so that empty username field shows an error
- Changed the submit button's label from "Edit Profile" to "Save Changes"
- Add a button to discard changes
- Edited delete team button style and removed the description since it is shown inside the modal

PS: I used hardcoded texts for some labels and Typography elements. This is because in my other PR #843, I handled all the hardcoded texts to be used from the translation files and created new keys for this file. I could not use those keys because the PR may not be merged. I also did not want to do the same steps twice not waste time. I will change the texts of EditTeam.jsx and EditTeamScripts.js as soon as that #843 is resolved.

## Changes regarding code refactoring

I noticed that the code in EditTeam.jsx and EditTeamScripts.js is copied from the EditProfile view and has many parts that should be changed or removed. These are the changes to simplify and refactor this code:

- EditProfile form's refs were deleted and two new refs for EditTeam form is created
- Unused functions and variables were removed
- In the old code, form values were saved in the state manually. Now, Formik's `handleChange` prop is used. Thus, the old state is removed.
- The handleSubmit function is moved to EditTeamScripts.js EditProfile function so that the form submits after all the validations run.
- New id's and names were given to the form elements because the existing id's and refs caused the bio field to get the data from the user's profile. 
- To prevent previous complications to happen, `getProfile()` function in the effect is removed because the code does not use any of its return values.
- Edited getTeamProfile function to fill the username and bio fields if they have an initial value
- Validations are added for the form elements

## The backend change

After using the Formik props, the form did not submit when the user left the bio field empty. In the frontend, the bio field is sent to the backend as empty string if the user sent it empty. The form worked as expected whwn the bio is filled. When I checked edit profile form's bio field, it is also sent as an empty string when submitted empty. Thus, I could not prevent this from happening with changes to the frontend. 

https://stackoverflow.com/a/55150513 => Assuming that the problem is caused because Django interprets empty string as falsy, I changed the line to raise a validation error only if the groupname is missing.

I am not really sure about this solution, looking forward to reviews 🙂

## Videos

Before

https://github.com/unstructuredstudio/zubhub/assets/92817363/2405c0b7-c382-4a49-b722-989052be7c4a

After

https://github.com/unstructuredstudio/zubhub/assets/92817363/209e3693-3a21-4207-af22-69a7ab9e2daa


brrkrmn:
For future improvement, some functions' names can be changed to be more appropriate for the view. Instead of editProfile => editTeam etc.

zubhub_frontend/zubhub/src/views/edit_team/EditTeam.jsx
```diff
@@ -308,47 +257,63 @@ function EditTeam(props) {
...
+                  <Grid item xs={12}>
+                    <Link to={`/teams/${groupname}`} className={classes.textDecorationNone}>
+                      <CustomButton
+                        variant="outlined"
+                        size="large"
+                        secondaryButtonStyle
+                        customButtonStyle
+                        fullWidth
+                      >
+                        {'Discard Changes'}
```
coderatomy: Consider translating this

zubhub_frontend/zubhub/src/views/edit_team/EditTeam.jsx
```diff
@@ -308,47 +257,63 @@ function EditTeam(props) {
...
+                    variant="outlined"
+                    size="large"
+                    secondaryButtonStyle
+                    dangerButtonStyle
+                    fullWidth
+                    onClick={() =>
+                      handleSetState(handleToggleDeleteAccountModal(state))
+                    }
+                  >
+                    {'Delete Team'}
```
coderatomy: same here

coderatomy (changes requested):
Nice work @brrkrmn. Just a few thing you missed out
Also consider resolving the conflicts

zubhub_frontend/zubhub/src/views/edit_team/EditTeam.jsx
brrkrmn: Actually I was waiting for another PR to be merged to translate the texts. That is merged, so I am working on translating these as well, thank you! 🙂

brrkrmn:
@shrish, I replaced the texts with translations. This is ready for review ☺️

tuxology:
@brrkrmn Please refer the Figma design to see edit team implementation:

image
 

You can use existing styles components from what is being done in #778 I agree to your other points about code refactoring, but lets take correct components from similar styled pages such as Create Team or from work done in #778. In addition, ensure this works in responsive mode as shown in screenshots

brrkrmn:
@tuxology Hey, I am currently working on this according to the Figma design. I see that editTeam.jsx's styles are imported from efitProfileStyles.js. Should I create a new editTeamStyles.js file to separate the view's design?

brrkrmn:
@tuxology, I've changed the files according to the Figma design. 
I created an editTeamStyles.js file, which is the duplicate of the existing styles except for the unused parts. I also added new classes for the dialog part. 
In the Figma design, there is not a submit button but the error handling and validation in this form requires one, so I kept the submit button. 
Here is the final version:
Screenshot 2023-10-16 at 17 48 04
Screenshot 2023-10-16 at 17 48 25

And the mobile version:
Screenshot 2023-10-16 at 17 47 33
Screenshot 2023-10-16 at 17 47 40

tuxology (approved):
LGTM 🌈

tuxology (changes requested):
Upon closer look, have a few comments: 

Is it possible to use https://github.com/unstructuredstudio/zubhub/blob/master/zubhub_frontend/zubhub/src/components/modals/Modal.jsx#L16 for your modal instead of creating new styles? If you would like to change the design, its fine, lets change it in a single place. I am re-marking the PR for changes.

brrkrmn:
@tuxology I used our Modal component for the dialog. I moved the dialog styles to the component's style file but did not style the component itself to avoid changing other modals in the app. This view now uses classes from the modal component, but for reusability, I think some changes can be made to the modal component in terms of styling.

tuxology (approved):

PR #868, opened 2023-10-11
feat: fixes the scrollable sidebar

### LINKED ISSUE #850: Make side navigation menu sticky
**Description**
The main functionality of zubhub is to create and share projects. To create a project we have to go to the sidenav. But since the sidenav goes away when we scroll up, there is no other way the user can create a project when they have scrolled all the way down to the projects.
Screenshot from 2023-10-09 16-47-12

Solution:
1. We either have a separate CTA button to create projects which is visible at all times. (This will require a lot of redesign, is not feasible as an immediate measure).
2. We fix the sidenav to the viewport. I have raised a PR for this in  #849 

I would love a designer's perspective on this. But I prefer the second option for now.

yokwejuste:
## Summary

Description of PR here...
Closes #850

## Changes

- Changes the position of the sidebar to `fix.

### DIFF
--- a/zubhub_frontend/zubhub/src/components/Sidenav/sidenav.styles.js
+++ b/zubhub_frontend/zubhub/src/components/Sidenav/sidenav.styles.js
@@ -6,18 +6,15 @@ export const sidenavStyle = theme => ({
         minHeight: '80vh',
         display: 'flex',
         borderRadius: 8,
-
         [theme.breakpoints.down('sm')]: {
             height: '100%',
         },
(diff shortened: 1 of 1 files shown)


tuxology:
@yokwejuste We are closing this since #849 has been reopened. We will let @aqsaaqeel work on it

yokwejuste:
> @yokwejuste We are closing this since #849 has been reopened. We will let @aqsaaqeel work on it

Alright

PR #870, opened 2023-10-11
(fix) Border radius in profile

brrkrmn:
## Summary

The right top and bottom border-radius of the about box were overridden by another border-radius value.

## Changes

- Deleted the later border-radius value and gave 8px border-radius for the left 

## Screenshots
Before

Screenshot 2023-10-11 at 18 07 10

After

Screenshot 2023-10-11 at 18 16 13


aqsaaqeel:
Hey @brrkrmn The two cards are supposed to be two different cards with space in between as we can see in the Figma designs. Refer : [this](https://www.figma.com/file/hidnBfP4kNWl7RDxy3aUZI/ZubHub?type=design&node-id=720-3287&mode=design&t=R3IHHToLof6geYvl-0) 
Screenshot from 2023-10-12 21-27-50

Also, I think we should stick to making an issue first before making a PR as the mentors have said. This would help the designers to also review the changes before we implement them easily.

cc: @srish @tuxology

brrkrmn:
@aqsaaqeel oh, thank you for reminding, my bad. I thought since this is a small fix there is no need to open an issue. By the way, these two badges are still separate in the mobile version thanks to this code:

```
    borderRadius: '8px 0 0 8px',
    [theme.breakpoints.down('900')]: {
      margin: '1em',
      borderRadius: '4px',
    },
```

I only changed how it looked in the above image when they are side by side. Here is the mobile version in my implementation just to make sure.

Screenshot 2023-10-12 at 20 11 48

tuxology (approved):

Issue #801, opened 2023-10-05
Improve UX of header in mobile screen


N-TITIAN:
I was going through the site and saw the contents of the header section is not properly aligned causing some inconsistency 

![image](https://github.com/unstructuredstudio/zubhub/assets/81820848/892230b7-ea1a-4671-8ab5-47c739c94373)
So i came up with design as a designer .  @srish @tuxology 
![Android Small - 1](https://github.com/unstructuredstudio/zubhub/assets/81820848/ef74b956-5ed6-4fb0-9e4e-0a8f11cbb71e)

srish:
@N-TITIAN I am not able to replicate this issue on my end. Could you please share which OS version and device you are seeing this issue with? Also, how is your design better than the existing one? I see no big difference here.

N-TITIAN:
Thank you @srish I saw the issue while using a smaller screen size (307x604) but viewing with a bigger mobile screen now ,the  alignments looks okay. I think its a minor issue with the sites responsiveness to smaller screens.

coderatomy:
Nice catch @N-TITIAN. Noticed the strange alignment of the Zubhub header and the internationalization icon. @srish, I believe we need a change for this for a better look on mobiles.

@N-TITIAN, it looked nice when they're all in-line. Maybe we reduce the font-sizes. What  are your suggestions?
![image](https://github.com/unstructuredstudio/zubhub/assets/141822315/c1c13d3f-1f87-4564-9f14-131de04a7295)

N-TITIAN:
@coderatomy I think maintaining them all in -line is a nice idea 
![Android Small - 2](https://github.com/unstructuredstudio/zubhub/assets/81820848/6cac02c1-8b49-4b03-8d2f-e3bfaf3d7500)

coderatomy:
Still too  much populated! How about we move the profile icon to the side nav only on mobile? Then we remain with only the notification and search icons. I think that's a better solution

Stephxdesigns:
@coderatomy here's something I came up with
![Screenshot_20231008-044659_1](https://github.com/unstructuredstudio/zubhub/assets/136924367/58a0e501-9834-451e-acd3-8277be32e454)
All the elements are aligned properly and it's not populated 

cc @srish @tuxology

N-TITIAN:
@coderatomy I looked into your suggestion and came up with this.
@srish @tuxology 
[...]

PR #883, opened 2023-10-12
fix: header styles and heading of projects

### LINKED ISSUE #792: Improve the header design on the landing page
Review@2x

1. The Hero Text lacks Hierarchy and the CTA button isn't bold enough to call attention to the user. 
2. The logo and hero text are not aligned vertically
3. The Creator Review section has a visibility issue. The text color and background color do not pass the contrast check.

aqsaaqeel:
## Summary

-> Used the finalized design in https://github.com/unstructuredstudio/zubhub/issues/792 to develop the hero section. 

-> The hero table is not populated so the image needs to be added from the backend. 

-> To run it and test, I added the Hero details on [http://localhost:8000/admin](https://github.com/unstructuredstudio/zubhub/pull/url)

-> As for this PR, I have made all the necessary CSS changes and added the necessary components.

## Changes

- Hero Container design on landing page
- Translated title for both English and Hindi
- Font Size of the heading Featured Project to match the rest

## Screenshots
I have made the changes according to @srish instructions [here](https://github.com/unstructuredstudio/zubhub/issues/792#issuecomment-1756589431)
Screenshot from 2023-10-12 18-43-00


aqsaaqeel:
I have made the changes you suggested @tuxology SInce the hero section text had a different style than the rest of the titles, they needed new classes for styles to be made. Here is how the hero section looks now.
Screenshot from 2023-10-13 10-44-13
Screenshot from 2023-10-13 10-43-38

zubhub_frontend/zubhub/src/views/home/Home.jsx
```diff
@@ -103,8 +102,8 @@ function Projects(props) {
           (
             <Box className={classes.root}>
               {hero && hero.id ? (
-                 <Box className={classes.heroSectionStyle}>
-                 <Box className={classes.heroContainerStyle}>
+                 <Box className={classes.heroMainSectionStyle}>
```
tuxology, line 105: Why have we renamed it to `heroMainSection` style instead of old name?

tuxology (changes requested):
Few changes we need: 
 - [ ]  The UI does not match the design in terms of width of hero element. It is currently full width however in the proposed design, we need to align it in the same width as the container containing elements of navbar. 
 -  [ ] In mobile view, there needs to be padding between Share your Work button and the image and the image needs to be center aligned

zubhub_frontend/zubhub/src/views/home/Home.jsx
aqsaaqeel: There is another section on this page that uses the ```heroSectionStyle``` name for its classes. It is the carousel section for ambassadors and their testimonials. 
Making any changes to the already existing classes meant changing other sections as collateral too. However, since we were focusing on just the Hero Section and it required different styling than the rest of the components, I made a new class.

aqsaaqeel:
> Few changes we need:
> 
> * [ ]   The UI does not match the design in terms of width of hero element. It is currently full width however in the proposed design, we need to align it in the same width as the container containing elements of navbar.
> * [ ]  In mobile view, there needs to be padding between Share your Work button and the image and the image needs to be center aligned

Changes I have made in this commit:
- The padding for `navbar` is adjusted.
- Padding space between `hero button` and `hero image` has been added.

Screenshot from 2023-10-14 05-54-03
As for the image on mobile screen, it looks centered here, is this what you meant? @tuxology 
Screenshot from 2023-10-14 05-51-56


Since the scope of this PR is just the hero section, I haven't altered much on the navbar, but I feel like we should also implement the design for the icons from [this](https://www.figma.com/file/nyUduEpFmVZ05xA8rt28Mt/ZubHub-Feature-Re-Designs?type=design&node-id=11-414&mode=design&t=xDeX7LSKbik6YPFy-0)

zubhub_frontend/zubhub/src/assets/js/styles/views/projects/projectsStyles.js
```diff
@@ -108,20 +137,44 @@ const styles = theme => ({
...
     textTransform: 'none',
-    padding: '0 1em',
-    fontSize: '1.2rem',
+    fontSize: '1.456rem',
     fontWeight: 700,
     marginRight: '1em',
     marginTop: '1.2em',
+    width: 'fit-content',
+    height: '3.95rem',
+    padding: '1.04rem 2.08rem 1.04rem 2.08rem',
```
tuxology: So there are two decimal places here. Why do we need such accuracy?

zubhub_frontend/zubhub/src/assets/js/styles/views/projects/projectsStyles.js
```diff
@@ -108,20 +137,44 @@ const styles = theme => ({
...
+    fontWeight: 500,
+    [theme.breakpoints.down('sm')]: {
+      fontSize: '1.2rem'
+    },
+  },
   heroButtonStyle: {
     textTransform: 'none',
-    padding: '0 1em',
-    fontSize: '1.2rem',
+    fontSize: '1.456rem',
```
tuxology: Why do we need 3 decimal accurate values? How have these been calculated? I feel if we need such fine adjustments, then there must be some underlying issue

tuxology (approved):
I am fine with this, but need a few clarifications/modifications on comments before we merge @aqsaaqeel

zubhub_frontend/zubhub/src/assets/js/styles/views/projects/projectsStyles.js
aqsaaqeel: I use the Figma design to show me the exact styles applied, and those are in `px`. I convert them to `rem` because it is more flexible in terms of adjusting according to screen sizes. The `px` values are a whole number but converting them to `rem` gives these decimal values.

zubhub_frontend/zubhub/src/assets/js/styles/views/projects/projectsStyles.js
aqsaaqeel: No, I don't think there is any underlying issue. My reasoning is the same as above ☝️

zubhub_frontend/zubhub/src/assets/js/styles/views/projects/projectsStyles.js
tuxology: @aqsaaqeel Can we round the double digit values to single digit at the very minimum? You can eyeball them to see if they align. Usually, there should be minimum need to do such "fine tuning". The figma designs are not supposed to be accurate representations to the exact location, but just indication of how things are aligned. Another benefit of keeping simple numbers is that we don't have to do a lot of fine tuning in other places (containers/padding) - since they will also align to single values.

aqsaaqeel: Alright. I'll make the changes!

tuxology:
@aqsaaqeel Can you revert navbar height? We don't want to change it yet

tuxology (changes requested):
See comment about navbar height.

aqsaaqeel:
> See comment about navbar height.

Alright! Can I work on #801 too on this PR? @tuxology

tuxology (approved):

tuxology:
@aqsaaqeel Lets leave #801 for now.

Also, sorry for increasing the scope of this PR, but can we actually harcode the header SVG here in the code itself? (add the  SVG to the PR and use it). We can keep the custom image upload feature (which is currently in Django Admin) still. We can add a checkbox in the Django Admin to select default image to be used OR uploaded image. Is that doable?

aqsaaqeel:
> We can add a checkbox in the Django Admin to select default image to be used OR uploaded image

It's alright. I'm on it!

aqsaaqeel:
@tuxology I've added the svg image but I am not able to understand why we need the optional image on admin. We are controlling what goes in the header image in the frontend anyway. So wouldn't the choice of which image we want to display come from the frontend even if we want to put the image from Admin? 🤔

zubhub_frontend/zubhub/src/assets/js/styles/views/projects/projectsStyles.js
```diff
@@ -25,20 +25,35 @@ const styles = theme => ({
...
   },
-  heroSectionStyle: {
+  heroMainSectionStyle: {
     backgroundColor: 'var(--primary-color1)',
-    boxShadow: '-2px 25px 13px -18px rgba(220,53,69,0.7)',
-    WebkitBoxShadow: '-2px 25px 13px -18px rgba(220,53,69,0.7)',
-    MozBoxShadow: '-2px 25px 13px -18px rgba(220,53,69,0.7)',
     display: 'flex',
     paddingTop: '2.3em',
     paddingBottom: '2.85em',
```
tuxology: Can you do rounding?

zubhub_frontend/zubhub/src/assets/js/styles/views/projects/projectsStyles.js
```diff
@@ -99,6 +127,7 @@ const styles = theme => ({
   heroMessageContainerStyle: {
     flexGrow: 1,
     color: 'white',
+    maxWidth: '37.813rem',
```
tuxology: Rounding

zubhub_frontend/zubhub/src/assets/js/styles/views/projects/projectsStyles.js
```diff
@@ -108,20 +137,44 @@ const styles = theme => ({
...
   heroButtonStyle: {
     textTransform: 'none',
-    padding: '0 1em',
-    fontSize: '1.2rem',
+    fontSize: '1.456rem',
     fontWeight: 700,
     marginRight: '1em',
     marginTop: '1.2em',
+    width: 'fit-content',
+    height: '3.95rem',
```
tuxology: Rounding

tuxology (changes requested):
Please review the code again for decimal rounding suggestions. Eyeball the UI once you have rounded the values.

tuxology:
> @tuxology I've added the svg image but I am not able to understand why we need the optional image on admin. We are controlling what goes in the header image in the frontend anyway. So wouldn't the choice of which image we want to display come from the frontend even if we want to put the image from Admin? 🤔

The idea for user provided hero element was to make the image modifiable from time to time (eg. highlight some news, upcoming event etc). The flow is this:
 - Default image is one hardcoded in frontend. In absence of any header image upload from Django Admin, this is what is shown
 - If custom image is uploaded AND the checkbox ticked, then the custom image takes precedence in rendering. 
 - If custom image is uploaded but the checkbox is not ticked, default image will be used

This flow can be coded in the frontend to get this working.

aqsaaqeel:
> > @tuxology I've added the svg image but I am not able to understand why we need the optional image on admin. We are controlling what goes in the header image in the frontend anyway. So wouldn't the choice of which image we want to display come from the frontend even if we want to put the image from Admin? 🤔
> 
> The idea for user provided hero element was to make the image modifiable from time to time (eg. highlight some news, upcoming event etc). The flow is this:
> 
> * Default image is one hardcoded in frontend. In absence of any header image upload from Django Admin, this is what is shown
> * If custom image is uploaded AND the checkbox ticked, then the custom image takes precedence in rendering.
> * If custom image is uploaded but the checkbox is not ticked, default image will be used
> 
> This flow can be coded in the frontend to get this working.

I have added the conditional to render the hardcoded image if there isn't one coming from the admin already.
As per the use case which you have mentioned, when we want a modified image, we can:
-> Go to admin panel.
-> Upload custom image.
After the said event expires, we can remove the custom image.
-> To clear it, there's already a clear selection box given. We can clear, save changes. And refresh.

Untitled design (1)



Do we have to add a checkbox on top of this ☝️
 
There is another issue that we can see now. The image coming from the admin works properly but the concern is the design does not have padding on the right as we can see here. (this is not an issue for the hardcoded image because of how it is designed) Should I work on this?

Screenshot from 2023-10-20 00-23-55

tuxology:
I think this is fine for now 👍 We can iteratively improve it further

tuxology (approved):
LGTM 🌈

Issue #781, opened 2023-10-04
Add pagination for projects & activities page


Temabo:
The navigation will be difficult when there will be too many items on the [projects](https://zubhub.unstructured.studio/projects) and [activities](https://zubhub.unstructured.studio/activities) page (which only users with staff credentials have access to). Right now it is possible to navigate by clicking on the "Next" and "Previous" buttons. Adding a pagination feature with numbered pages will significantly improve user experience, making it much easier for users to move around and find the projects and activities they want. This small change will enhance the user-friendliness and efficiency of ZubHub.

aqsaaqeel:
Nice idea! I'd like to work with @Temabo to implement this feature @srish @tuxology

srish:
@Temabo, feel free to go ahead and create a design. Once it's ready, you can post it here so that we can gather input from others, and then @aqsaaqeel can put it into action. This aligns perfectly with our vision for how the designer and developer will work together on Outreachy microtasks. :-)

Temabo:
Great! @srish I will start working on it and tag @aqsaaqeel to the figma link when I am done

Temabo:
Hello @aqsaaqeel This is the link to the figma file that contains the pagination design.
https://www.figma.com/file/A1GYQtENXSmPUzB63MTgPj/outreachy?type=design&node-id=506%3A7405&mode=design&t=xdOBknDLPUnlp7LB-1
![Pagination](https://github.com/unstructuredstudio/zubhub/assets/105416140/926501e3-cd19-4dfd-84d3-d1cfe6d02715)

PascalAnuche:
With the pagination concept, @Temabo  came up with, I also came up with an idea of how the layout can be when users are navigating through the featured projects and All Projects. The pagination will help the users to navigate through the featured projects easily by seeing 6 All projects horizontally decreasing the length of the page which helps in a seamless user experience and saves time too.

Here is the figma link https://www.figma.com/file/zWHBhu5EnsNJbMi26j1NKb/Outreachy-Zubhub-intern?type=design&node-id=0%3A1&mode=design&t=BPYegXaKZF0212Vm-1

@srish @Temabo 
<img width="1622" alt="Frame 427321225" src="https://github.com/unstructuredstudio/zubhub/assets/101313594/b2ae9d38-be54-44be-86b6-159b76702cf9">

chudivictor:
[...]

PR #888, opened 2023-10-12
(Feat) scroll to top when changing project page

### LINKED ISSUE #743: Scroll to top not implemented when new projects are listed
**Describe the bug**
When you login and go to the `all projects` link, you notice a list of projects. As you scroll to the end and reach the last one on the screen with the `next` button, clicking on the next button would fetch a new list of projects. The bug here is that the page doesn't automatically scroll to the top. Instead the user has to manually scroll to the top of the screen to find the newly listed projects.

**To Reproduce**

1. Go to 'All Projects `https://zubhub.unstructured.studio/projects`'.
2. Click on `'next'` at the bottom of the screen.
3. You notice the old list disappears but you have to manually scroll to the top. 

**Expected behavior**
A scroll to top action should automatically occur when the new list is fetched successfully.

brrkrmn:
## Summary

When changing the page of the projects in https://zubhub.unstructured.studio/projects, the user is not directed to the top of the page.
Closes #743

## Changes

- Added `window.scrollTo(0, 0)` to the onClick functions of the prev and next buttons

##Before

https://github.com/unstructuredstudio/zubhub/assets/92817363/e29577e8-8ca2-49a9-a899-eadf99c35cc5

##After

https://github.com/unstructuredstudio/zubhub/assets/92817363/0d7c6991-e5f7-45c7-be5f-ec7f84ee8d52


rak16:
Nice 👍

tuxology:
@brrkrmn I recall implementing scroll to top and added it application wide: https://github.com/unstructuredstudio/zubhub/pull/725/files#diff-eccfbd0e78d8ecab6691df1fff5f1d743db1359b05475d56c37b95c625554e59 Can you debug why its failing here or reuse that?

brrkrmn:
> @brrkrmn I recall implementing scroll to top and added it application wide: https://github.com/unstructuredstudio/zubhub/pull/725/files#diff-eccfbd0e78d8ecab6691df1fff5f1d743db1359b05475d56c37b95c625554e59 Can you debug why its failing here or reuse that?

Hey, I think I figured what's wrong here. `ScrollToTop` runs only when the `pathname` coming from `useLocation` changes. So, whenever the route changes, we scroll to top. In the projects page, going to the next or previous page does not change the route at all, thus, ScrollToTop does not run. 

I haven't found a way to trick this logic and simulate a route change but I do not think that would be a good idea because this situation creates other problems as well. For example when the user visits the second page of projects and refreshes the page, they are automatically redirected to the first page, which is rendered in /projects by default.

I believe it is best to actually change the route when we change project page like: /projects/page/2. This PR can be temporarily useful, but we can wait for #781 to be concluded to permanently solve this issue.

tuxology (approved):

PR #889, opened 2023-10-13
feat: added pagination to projects page

### LINKED ISSUE #781: Add pagination for projects & activities page

aqsaaqeel:
## Summary

Added pagination to projects page. 

## Closes #781 

Based on consensus, [this design](https://github.com/unstructuredstudio/zubhub/issues/781#issuecomment-1747819610) by [@temabo](https://github.com/Temabo) is what I implemented.

## Changes

- Added MUI Pagination Comp on projects page
- Added a custom ```page_size``` parameter to pagination feature so now we can configure the number of projects displayed using the url in frontend.
- Added css classes to give custom style to pagination comp.

## Screenshots
Here's how it looks on different screen size and different number of projects displayed.
Screenshot from 2023-10-13 06-08-40
Screenshot from 2023-10-13 05-31-11
Screenshot from 2023-10-13 05-30-50
Screenshot from 2023-10-13 06-27-24


tuxology:
Nice attempt @aqsaaqeel and @temabo 🎉 I'll review this!

zubhub_frontend/zubhub/src/api/api.js
```diff
@@ -854,7 +854,7 @@ class API {
    * @todo - describe method's signature
    */
   getProjects = ({ token, page }) => {
-    const url = page ? `projects/?${page}` : `projects/`;
+    const url = page ? `projects/?page=${page}&page_size=${PROJECTS_PAGE_SIZE}` : `projects/?page_size=${PROJECTS_PAGE_SIZE}`;
```
tuxology: OK, so I don't understand the goal of using `PROJECT_PAGE_SIZE` here in URI. This is mostly relevant for getting total number of pages you have to show in the pagination view isn't it? In that case, we can just decide on a fixed number, like 18 and put it in utils.js. Then in UI, we can adapt on how many paginations are we showing. eg

<---- [1, 2. 3, 4, 5, 6... ] of [256]   --->

As you reach number 6, the view should then auto adjust UI as:

<--- [1...6, 7, 8, 9, 10] of [256] ---> 

and so on.

tuxology (commented):
I also think that we need to ensure all cases are working. 0 projects, 1382 projects, 2 projects. Can we do some synthetic testing to ensure that the pagination works on all cases?

zubhub_frontend/zubhub/src/views/projects/Projects.jsx
```diff
@@ -57,10 +60,11 @@ function Projects(props) {
   };
 
   const { loading } = state;
-  const { results: projects, previous: prev_page, next: next_page } = props.projects.all_projects;
+  const { count: totalProjects, results: projects, previous: prev_page, next: next_page } = props.projects.all_projects;
   const { hero } = props.projects;
   const staff_picks = props.projects.staff_picks;
   const { t } = props;
+  const noOfPage = Math.ceil(totalProjects / PROJECTS_PAGE_SIZE) ;
```
tuxology: Is `totalProjects` here total projects including drafts (which are not visible)? In that case, pagination calculation will be off

zubhub_frontend/zubhub/src/api/api.js
aqsaaqeel: > This is mostly relevant for getting total number of pages you have to show in the pagination view isn't it?

Yes!

> In that case, we can just decide on a fixed number, like 18 and put it in utils.js.

The backend has a hard-coded value of `18` right now which will be treated as the default if we don't pass the `page_size` param. We could hard-code it in the frontend too like you suggest. But the reason I added this is to allow flexibility on the frontend side. Say we ever want to change the number of projects per page in the future; all we have to do is change this configuration on the frontend and it will just work. No need to make any changes on the backend at all. If we go with the hard-coded approach, we'd have to make changes on both ends. 

> In that case, we can just decide on a fixed number, like 18 and put it in utils.js. Then in UI, we can adapt on how many paginations are we showing. eg
> 
> <---- [1, 2. 3, 4, 5, 6... ] of [256] --->
> 
> As you reach number 6, the view should then auto adjust UI as:
> 
> <--- [1...6, 7, 8, 9, 10] of [256] --->
> 
> and so on.
> 
And yes, this one is already taken care of.

zubhub_frontend/zubhub/src/views/projects/Projects.jsx
aqsaaqeel: > Is `totalProjects` here total projects including drafts (which are not visible)? In that case, pagination calculation will be off

It does not include drafts

zubhub_frontend/zubhub/src/api/api.js
kamthamc: If we accept the page size parameter we also have to validate the minimum and maximum values in the backend

zubhub_frontend/zubhub/src/assets/js/styles/views/projects/projectsStyles.js
```diff
@@ -271,11 +271,30 @@ const styles = theme => ({
...
+    paddingLeft: '2em',
+    paddingRight: '2em',
+    display: 'flex',
+    justifyContent: 'space-between',
+    marginTop: '2em',
+  },
+  paginationRoot: {
+    '& .MuiPaginationItem-root': {
+      '&:hover': {
+        backgroundColor: 'var(--primary-color3)',
```
tuxology, line 307: 👍

zubhub_frontend/zubhub/src/views/projects/Projects.jsx
```diff
@@ -164,35 +168,48 @@ function Projects(props) {
...
                       handleSetState({ loading: true });
                       handleSetState(fetchPage(page, props));
                     }}
                     primaryButtonStyle
                   >
                     {t('projects.prev')}
                   </CustomButton>
-                ) : null}
+                ) : <CustomButton className={clsx(classes.visibilityNone)}></CustomButton>}
+                    <Pagination
```
tuxology: Can we align this horizontally centered with the Next and Prev button. We can see in screenshots that this is not

zubhub_frontend/zubhub/src/api/api.js
tuxology: @aqsaaqeel I feel we should fix the page size parameter to 18 for now and not allow users to set it. We can introduce user configurable values later. For now, we want to keep it simple and navigable

tuxology (changes requested):
A few more changes needed

aqsaaqeel:
@tuxology Here's how it looks now with:

1. 1382 projects

Screenshot from 2023-10-18 03-29-04

2. 54 projects

Screenshot from 2023-10-18 07-08-19

3. 2 projects

Screenshot from 2023-10-18 07-00-10

4. 0 projects

The default error page with the Create Project button

5. On Mobile Screen (There wasn't a specific design for mobile so I took creative liberty and removed "Next" and "Prev" text on buttons)

Screenshot from 2023-10-18 07-08-33

tuxology:
Very well done and tested @aqsaaqeel !  🎉 So this shows us one potential issue. When we have less than 1 page worth of projects, we just see  [1] there. We should now show pagination if content fits one page isn't it?  Lets cater to this case and we can then merge!

aqsaaqeel:
> Very well done and tested @aqsaaqeel ! 🎉 So this shows us one potential issue. When we have less than 1 page worth of projects, we just see [1] there. We should now show pagination if content fits one page isn't it? Lets cater to this case and we can then merge!

Thank you. It now works for `lesser < 19 projects` case too. We do not see any pagination component when projects are less than 19 in number @tuxology

aqsaaqeel:
@tuxology this is ready to merge

tuxology (approved):

PR #891, opened 2023-10-13
Revert "fix: fix sidenav to view port (#849)"

tuxology:
Reverting this since we need a better solution than just doing position: 'fixed'

image


Mitchell377:
@tuxology  what do you want to be implemented from the design perspective?, please let me know

PR #893, opened 2023-10-13
feat: add additional password validation on Signup

aqsaaqeel:
## Summary

Requiring strong passwords on all websites is a standard practice for enhancing security. Currently, Zubhub only has password validation on signup for passwords based on length. I have added password validation on signup for.

- At least 1 upper-case character.
- At least 1 lower-case character.
- At least 1 digit.
- At least one special character ("!", "@", "#", "$", "%", "^", "&", "*")

## Screenshots

Screenshot from 2023-10-13 19-58-07
Screenshot from 2023-10-13 19-36-27
Screenshot from 2023-10-13 20-00-08
Screenshot from 2023-10-13 19-37-05

### DIFF
--- a/zubhub_frontend/zubhub/public/locales/en/translation.json
+++ b/zubhub_frontend/zubhub/public/locales/en/translation.json
@@ -117,7 +117,7 @@
           "min": "Your password seems too short",
           "upper-caps": "Your password must contain at least 1 uppercase character",
           "lower-caps": "Your password must contain at least 1 lowercase character",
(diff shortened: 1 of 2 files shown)


tuxology:
Passwords length is more important than adding complex characters - hence we have not added any restriction for special characters. Having less complex characters is also important because this is directed towards young audience. We will increase minimum length soon if neeed - rest other changes are not needed. I am closing this one. In addition, please create issues so we can discuss them first. Thanks 👍

PR #908, opened 2023-10-15
feat: added placeholder for activity logs

### LINKED ISSUE #907: fix: Show message when there is no activity on Profile Page
**Description**
Currently, it does not show any activity under **Activity Log** in the profile for new users. Here's how it looks.

Screenshot from 2023-10-16 00-00-24

As we can see, even though there is no badge for new users still there's some text that tells the user what to expect. We should do the same and add a text when there is no activity yet. We can write this text "Seems like there isn't any activity on your account yet! Get involved to see activity logs!"

Screenshot from 2023-10-15 08-52-06

aqsaaqeel:
## Summary

I added some placeholders for activity logs for new users. I noticed there is placeholder text for badges for new users but none for Activity Logs

Closes #907 

## Currently Looks Like 

Screenshot from 2023-10-16 00-00-24

## After Change

Screenshot from 2023-10-15 08-52-06


zubhub_frontend/zubhub/src/views/profile/Profile.jsx
```diff
@@ -304,12 +303,13 @@ function Profile(props) {
...
 
                     {
+                      userActivity.length ? 
                       userActivity.map(activity => (
                         <UserActivitylog 
                         activity={activity}
                         key={activity.id}
                         />
-                        ))
+                        )) : (<Box>Seems like there isn't any activity on your account yet! Get involved to see activity logs!</Box>)
```
tuxology: Use translation string

tuxology (changes requested):

zubhub_frontend/zubhub/src/views/profile/Profile.jsx
aqsaaqeel: Alright

tuxology (approved):

PR #911, opened 2023-10-16
None - removed from the Project & Activity Section

### LINKED ISSUE #875: None option is not available in creating a project and activity
Recently, I tried to create a project then when I got to the last step which is Step 3  I was asked to select none if I am unsure of the category.

Looking at this I suggest that a none  checkbox which disables all other checkbox when selected should be provided as an error prevention feature and to also provide easy understanding and navigation for the user
See screenshot below  and please can I work on this issue as a designer @srish 

Screenshot 2023-10-12 064537

Urmilasc:
**Description : ** 
This PR removes this line " Select none if you are unsure about your category." from both the Project and Activity Section

Closes #875 

**Changes :**

Removed the line " Select none if you are unsure about your category." from the create Project and Activity Create Area.

**Steps to test : **

- Login to Zuphub
- Click on the create project present on the side navbar 
- After clicking on that , start to create a project and go upto 3rd step , there the line of which come as an helperText has been removed .

**Screenshot after change : **

helper

Please review it and let me know the changes to be done !!

### DIFF
--- a/zubhub_frontend/zubhub/src/views/create_activity/step1/Step1.jsx
+++ b/zubhub_frontend/zubhub/src/views/create_activity/step1/Step1.jsx
@@ -37,7 +37,7 @@ export default function Step1({ formik, ...props }) {
       <SelectFromPills
         name="category"
         label="What category does your activity belong too? "
-        helperText="Select three categories that best describe your project. Select none if you are unsure about your category."
+        helperText="Select three categories that best describe your project."
         data={categories}
         onChange={_.debounce(data => formik.setFieldValue('category', data), 200)}
         selectedItems={formik.values.category}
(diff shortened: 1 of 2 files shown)


aqsaaqeel:
We should add a none option too in case the user does not want to put it in a category.

Urmilasc:
Okay @aqsaaqeel , Thankyou for confirming , try to implement that one !!

tuxology:
> We should add a none option too in case the user does not want to put it in a category.

@aqsaaqeel I think in backend we don't have a None category (I may be wrong). It is ideal just to let user select any of the category, hence what @Urmilasc did is ok. We however need to pass this text via translation

tuxology (changes requested):
Use translation string for this

srish (approved):

tuxology (approved):
Actually its fine, we can merge this

PR #873, opened 2023-10-12
fix: adds a margin to the items

yokwejuste:
## Summary

Description of PR here...
Closes #866

## Changes

- The search result files

## Screenshots

image

PR #914, opened 2023-10-17
fix: spacing issue of search results

### LINKED ISSUE #866: No horizontal spacing between cards on search results page
**Describe the bug**
Search result for logged in users is not adequately spaced

**To Reproduce**

1. Visit 'Zubhub website'
2. Click on 'Log in and make sure you are logged in'
3. Search for project 'light'

**Expected behavior**
adequate spacing between cards

**Screenshots**
space

brrkrmn:
## Summary

The search results had no horizontal space between them. Considering that the search results should look like what we have in the all-projects view, I compared the two files and their styles. Realized that the rendering of the projects was exactly the same which is the `Project` component inside a `Grid` item. The only difference was that in the all-projects view, the results were wrapped in a `Grid` container which has `spacing={3}`. I applied the same to the search results and there was no need to manually add `marginBottom` to them so I removed the projectGridStyle class. Now the results use the same spacing as the all-projects view.

Closes #866 

## Changes

- Wrap the search results' grid inside a container grid with spacing.
- Remove projectGridStyle class

## Screenshots

Screenshot 2023-10-17 at 08 51 41


tuxology:
@brrkrmn Thanks. Since this was fixed by #873 we are closing this. However, i see the inspired approach that we merged. You can consider this as a contribution. Thanks 🎉 I left a comment https://github.com/unstructuredstudio/zubhub/issues/866#issuecomment-1769740467 for more suggestions on how to handle such cases.

PR #916, opened 2023-10-17
style: changed the design of creator tag in profile

### LINKED ISSUE #897: Improve design of creator status on the profile page
Hello @srish I noticed on my profile page that the creator status looks more like a an active button which shouldn't be so
Here is a screenshot of how it currently looks.
image

Here is another screenshot of an improved version on how it should look like. What do you think?
image

brrkrmn:
## Summary

Edited tag style as discussed,

Closes #897

## Changes

- Changed the styling of tags in `profileStyles.js`. I did not change the common class in `index.js` since many places used that and would be affected by this change. This change only affects profile and team tags since `Team.jsx's` styles are from `profileStyles.js`.

## Screenshots
Before:
Screenshot 2023-10-17 at 16 24 29
After:
Screenshot 2023-10-17 at 16 26 04
Screenshot 2023-10-17 at 16 24 56


tuxology:
@brrkrmn We will wait for better styles in #897 before you can continue here.

NdibeRaymond:
@brrkrmn any update on this? do we now have the better design like Suchakra suggested?

brrkrmn:
> @brrkrmn any update on this? do we now have the better design like Suchakra suggested?

No new design has been proposed for this

PR #927, opened 2023-10-19
(fix) Fix regression for #825

coderatomy:
## Summary
fixes regression for https://github.com/unstructuredstudio/zubhub/pull/825


tuxology (approved):
I was specifically looking for this. This is 2nd time this regression happened 😅 I guess I also missed it!

PR #946, opened 2023-10-23
Fix: Show user's avatar on navbar when logged in

### LINKED ISSUE #945: Navbar does not show user's avatar when logged in
**Describe the bug**
Right now there is a placeholder avatar on the navbar. It does not display the user's avatar even when logged in.

**To Reproduce**

1. Login on Zubhub.
2. Notice the avatar icon

**Expected behavior**
The avatar icon must show the user's avatar when the user is logged in. But it shows a placeholder icon instead.

**Screenshots**

Screenshot from 2023-10-23 20-12-23

We must see this avatar on the navbar too

Screenshot from 2023-10-23 20-14-32

aqsaaqeel:
## Summary

When logged in, the user's avatar was not showing up on navbar.

## Changes

- Added the avatar on Navbar.
- Gave the Avatar a white background to improve visibility.
- Removed the inline style of Avatar to maintain code uniformity.

## Screenshots
I added the avatar to Navbar.

Screenshot from 2023-10-23 20-24-46

However, I noticed that the avatar was low on visibility because of the navbar's red background. So I added a background color of white and it looks much better now.

Screenshot from 2023-10-23 20-43-35

### DIFF
--- a/zubhub_frontend/zubhub/src/assets/js/styles/index.js
+++ b/zubhub_frontend/zubhub/src/assets/js/styles/index.js
@@ -7,7 +7,9 @@ const styles = theme => ({
     borderRadius: 20,
     display: 'flex',
     justifyContent: 'center',
-    alignItems: 'center'
+    alignItems: 'center', 
+    cursor: 'pointer',
+    backgroundColor: 'white',
   },
   flexColumn: {
     flexDirection: 'column'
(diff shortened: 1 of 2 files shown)


tuxology (approved):
Nice 👍

rak16:
Good one 👍

PR #467, opened 2022-05-09
Refactor recommendation bar

Deepanshu039:
## Summary
refactorization of the recommendation bar to make it more informative with an enhanced UI.

Closes #419

## Screenshots

Screenshot (115)

Screenshot (91)

PR #949, opened 2023-10-24
(fix): search option text shrinking in autocomplete

### LINKED ISSUE #419: Search autocomplete option's main text shrinks before short text

brrkrmn:
The project title in the autocomplete options shrinks, making it hard to identify the project. Following the approach in #467 , I made changes to the option styles so that the title is visible in all screens.

Closes #419 

## Changes

- In NavBar.jsx, when searching for projects, the options are filtered to only show the logged-in user's projects. The filtering of the option is removed to show the projects from other users. (I need comments on whether this was a feature or a bug)
- The image is placed in the left side of the option
- the username is next to title if there is enough space, and under it when the title is long (this can be changed to make username next to title for big screens, and under it below a certain size, waiting for comments on this)
- In mobile screens, the long title is clamped into 2 lines to maintain its text to an extent

## Screenshots
Here is the 3 different sizes: 
Screenshot 2023-10-24 at 12 42 03
Screenshot 2023-10-24 at 12 41 54
Screenshot 2023-10-24 at 12 41 42

And here is the options with different title and username length: 
Screenshot 2023-10-24 at 12 42 25

Regarding the username length and other possible placements of it, I encourage you to check [my comment](https://github.com/unstructuredstudio/zubhub/issues/419#issuecomment-1772265140)


zubhub_frontend/zubhub/src/components/Navbar/Navbar.jsx
```diff
@@ -113,7 +113,7 @@ function PageWrapper(props) {
         } else if (searchType === SearchType.PROJECTS) {
           completions = await api.autocompleteProjects({ query, token });
           completions = completions
-          .filter(c=>( c.creator.id === props.auth.id && c.publish.type !== 1 ))
+          // .filter(c=>( c.creator.id === props.auth.id && c.publish.type !== 1 ))
```
brrkrmn: I will delete this line if there is no need to filter the projects and show all of them

zubhub_frontend/zubhub/src/assets/js/styles/components/autocomplete/optionStyles.js
```diff
@@ -1,50 +1,83 @@
...
   infoWrapper: {
     display: 'flex',
-    flexFlow: 'row nowrap',
     alignItems: 'center',
-    maxWidth: '50%',
+    flexFlow: 'row',
+    flexGrow: 1,
+    overflow: 'hidden',
+    textOverflow: 'ellipsis',
+    //first approach
```
brrkrmn: the first approach was keeping the username at right for big screens. You can see the comparison [here](https://github.com/unstructuredstudio/zubhub/issues/419#issuecomment-1772265140)

NdibeRaymond:
@brrkrmn I left some comments for you on the main issue

zubhub_frontend/zubhub/src/components/Navbar/Navbar.jsx
tuxology: We should show all projects here. I think we should remove this line and test if this works fine. This could be a regressions as Raymond suggests here: https://github.com/unstructuredstudio/zubhub/issues/419#issuecomment-1806989291

zubhub_frontend/zubhub/src/assets/js/styles/components/autocomplete/optionStyles.js
tuxology: Lets mover username below. I also wanted to highlight that for smaller screens where the title gets split in two lines and we get one more username line below, we need to align all this within the height of the image so everything looks proportional. At this time, it is bleeding from top and bottom a bit. Take design help if needed.

brrkrmn:
@tuxology pushed the changes that you mentioned above: 
- username is now below title,
- image height is aligned with the information for smaller screens
- fixed image stretching 
- username is handled in 1 line in case it is long

PR #973, opened 2023-10-26
fix: prevent project image from stretching

### LINKED ISSUE #952: Project image is stretched out of proportion
**Describe the bug**
The project cover image gets stretched out of the correct proportion, distorting the image dimensions. This can give a bad impression.

**To Reproduce**

1. Login
2. Click on any of your created project.
3. See error

image

**Expected behavior**
The cover image should not be stretched . A suggested solution would be to set the cover image to "cover" or "contain" rather than "fill". This should maintain the correct aspect ratio while filling the space provided for the image

**Screenshots**
When set to "contain":
image

When set to "fill":
image

I believe that with this, the display of projects will be rather appealing and maintain the dignity of the project.

brrkrmn:
In project details, the project image used to stretch to fit in the container, this PR addresses that.

Closes #952 

## Changes

- Give the container max height and width 
- Center the image in the container
- Use object-fit to prevent stretching

## Screenshots
Screenshot 2023-10-25 at 13 25 59
Screenshot 2023-10-25 at 13 26 55
Screenshot 2023-10-25 at 13 27 08
Screenshot 2023-10-25 at 13 25 49


zubhub_frontend/zubhub/src/assets/js/styles/views/project_details/projectDetailsStyles.js
```diff
@@ -106,28 +106,22 @@ const styles = theme => ({
...
-    left: 0,
+    top: '50%',
+    left: '50%',
     width: '100%',
     height: '100%',
-    [theme.breakpoints.down('959')]: {
-      width: '100%',
-      height: '100%',
-    },
+    transform: 'translate(-50%, -50%)',
```
tuxology: This will essentially ensure image dimensions remain consistent, but we see black bars always (since no image will always have exact same width x height ratio). I suggest we explore "zoom and fill" technique rather than this. Currently, we have a "stretch and fill"

tuxology (changes requested):

zubhub_frontend/zubhub/src/assets/js/styles/views/project_details/projectDetailsStyles.js
brrkrmn: If I understood correctly, we should zoom the image to fill the container rather than showing bars around it right? Changing objectFit to 'cover' achieves this perfectly, which works well with most of the images other than very long or wide ones. In that case, only some part of the image is seen. 

here is a long image with what we currently have (stretched):
Screenshot 2023-11-12 at 13 22 31

here is the zoomed version of the long image:
Screenshot 2023-11-12 at 13 21 12

and here is my first implementation using objectFit: contain: 
Screenshot 2023-11-12 at 13 45 11

the third option may be unnecessary for other images but it allows us to show the whole content to the user. What do you think of using this option without the black background? This way the user sees the whole image and there won't be unnecessary black bars for each image.

Or we can use 'cover' and 'contain' conditionally depending on the image ratio.

tuxology: @brrkrmn I think lets do the zoom version. So third version can be removed. We already allow users to click the iamge and see a full screen original aspect ratio version.

brrkrmn: thanks @tuxology, made the last changes to use the zoom version, this is ready now 👍🏻

tuxology (approved):

PR #989, opened 2023-11-02
(fix): improve notification panel 

### LINKED ISSUE #988: Scrolling on Notification Panel
**Describe the bug**
Hi @srish I noticed the notifications panel scrolls in both vertical and horizontal directions. Users may find it confusing and challenging to navigate through a panel that moves in two directions. It introduces unnecessary complexity and can lead to frustration. I suggest we just make it scroll vertically to improve the user experience. Vertical scrolling is a standard interaction in most digital interfaces, ingrained in user behavior. Introducing horizontal scrolling disrupts this familiar pattern and can be disorienting for users.

Also, the unread button on the panel looks disabled and not clickable.

**To Reproduce**

1. Go to your dashboard
2. Click on the notification icon
3. See error

**Screenshots**
Screenshot (77)

To solve this problem, I designed an improvement attached below

Frame 1000002850

The second issue is when we get a new notification, the notification bell stays the same and users are not notified. If the bell doesn't alert the user, they may miss out on important updates or messages.

I have designed a solution to this issue below

Header Sign Up Option

brrkrmn:
## Summary

Notification component's width was set to '340px' above a certain screen size, causing a horizontal scrollbar to appear for those sizes. 

Closes #988 

## Changes

- notificationStyle's width is always 100%
- removed left padding to align the notification with title
- added new width for the view dot to prevent shrinking when notificationLink's maxWidth is '365px'

### Before

https://github.com/unstructuredstudio/zubhub/assets/92817363/58ca0875-44cb-415e-9245-cad3f0479190

### After

https://github.com/unstructuredstudio/zubhub/assets/92817363/43fadecd-bc22-4d0f-aa5d-81adf27e3db0

## Update

- Edited popper and its arrow to be aligned with the buttons
Screenshot 2023-11-06 at 17 54 05

- Centered the content in the logged-out notification panel
Screenshot 2023-11-06 at 17 55 04

- Customize the scrollbar in panel 
- Edit panel button's not selected views 
- Added translations for all notification panel texts

https://github.com/unstructuredstudio/zubhub/assets/92817363/ec6c9f57-c1e7-4a57-8738-0e82d7503a30

### PS
Notification panel is responsive and works in mobile right now, but the notification and the signup panel that we use in home page (not logged in) does not work under certain screen size. Some other things also don't work in that page so I did not involve fixing them in this PR


NdibeRaymond:
hello @brrkrmn, thanks for working on this! it looks so much better than it was. I have a few thoughts on things we can improve on since we are working on this now:
* scrollbar should be at the edge of the notification popup. something like this (but with a better looking scrollbar):
zubhub_notif_improve
* it will help if you can squash your commits and reduce it to just one commit

brrkrmn:
thanks, @NdibeRaymond. Sure, I can do that 👍🏻

mehreeee:
Looks better now @brrkrmn nice job👍
What about including the ‘close’ icon so users can easily close the panel instead of clicking outside it? I feel this will be a lot easier and better?

tuxology (changes requested):
I see the concern regarding responsiveness that you have mentioned. We can file separate issue and fix this. This is fine for now. There is one hardcoded value of color that should be a variable or should be taken from the theme. I recommend fixing that. Apart from that we can merge this

brrkrmn:
thank you @mehreeee, close button sounds good, but I'm not sure how would it look like on the notification and signup panels when the user's not signed in. Do you think we should use it only for the logged in user's notification panel?

tuxology (approved):
LGTM 🌈

PR #993, opened 2023-11-08
(feat) Notification bell doesn't show notifications

### LINKED ISSUE #992: Notification bell doesn't show notifications
**Describe the bug**

Hi @srish @tuxology when we get a new notification, the notification bell stays the same and users are not notified. If the bell doesn't alert the user, they may miss out on important updates or messages.

**To Reproduce**

1. Go to your dashboard
2. Create a sample project
3. Clap or bookmark the project
4. Click the notification icon, and see new notifications, but no alert on the bell.

**Expected behavior**
The notification bell is supposed to alert the user of every new notification.

I have designed a solution to the problem below

Header Sign Up Option

coderatomy:
## Summary
Implements notifiers in case of numbers.
Closes #992 

## Changes
- Adds notifications number

## Screenshots
image
Screenshot from 2023-11-09 02-02-02

### DIFF
--- a/zubhub_frontend/zubhub/src/components/notification_button/NotificationButton.jsx
+++ b/zubhub_frontend/zubhub/src/components/notification_button/NotificationButton.jsx
@@ -1,21 +1,44 @@
 import ClickAwayListener from '@material-ui/core/ClickAwayListener';
-import React, { useRef, useState } from 'react';
+import React, { useRef, useState, useEffect, useMemo } from 'react';
 import styles from '../../assets/js/styles/components/notification_button/notificationButtonStyles';
 import commonStyles from '../../assets/js/styles/index';
 import NotificationPanel from '../notification_panel/NotificationPanel';
 
 import { makeStyles } from '@material-ui/core/styles';
 import { Notifications } from '@material-ui/icons';
+import API from '../../api/api'
 import clsx from 'clsx';
 import { colors } from '../../assets/js/colors';
 import HamburgerMenu from '../hamburger_menu/HamburgerMenu';
 const useStyles = makeStyles(styles);
 
-const NotificationButton = ({ className, notif }) => {
+const notificationSort = (a, b) => {
+  return new Date(b.date) - new Date(a.date);
+};
+
+const NotificationButton = ({ className, notif, auth }) => {
   const classes = useStyles();
   const commonClasses = makeStyles(commonStyles)();
   const [dropdownOpen, setDropdownOpen] = useState(false);
+  const [notifications, setNotifications] = useState([]);
   const buttonRef = useRef();
+
+  const unreadNotifications = useMemo(
+    () =>
+      Object.values(notifications)
+        .sort(notificationSort)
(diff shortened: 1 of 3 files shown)


tuxology:
@coderatomy Thanks for this. The current design does not match exactly the suggested UX in #992 (notification circle should be half outside the main bell circle). I also think we should experiment with increasing the font size of the notification count. In addition, have you tested the behavior when all notifications are read already? Does the icon work as expected (it should not show 0, but should just disappear).

tuxology (changes requested):
See comments

coderatomy:
Everything works as expected. For the design @mehreeee updated it to this

mehreeee:
> Everything works as expected. For the design @mehreeee updated it to this

hi @tuxology I iterated and updated the design to this current one. Maybe we could try increasing the font for greater legibility.

zubhub_frontend/zubhub/src/components/notification_button/NotificationButton.jsx
```diff
@@ -1,32 +1,62 @@
...
+
   return (
     <ClickAwayListener onClickAway={() => setDropdownOpen(false)}>
       <div>
         <div
           onClick={() => setDropdownOpen(!dropdownOpen)}
           ref={buttonRef}
           className={clsx(classes.notification, commonClasses.iconBox)}
-          style={{ cursor: 'pointer' }}
+          style={{ position: 'relative', cursor: 'pointer' }}
```
NdibeRaymond, line 50: can you help move this to notificationButtonStyles.js?

zubhub_frontend/zubhub/src/components/notification_button/NotificationButton.jsx
```diff
@@ -1,32 +1,62 @@
...
         >
-          <Notifications style={{ color: colors.primary, fontSize: 20 }} />
+        {
+          unreadNotifications?.length > 0 && (
+           <span className={classes.spanNumber}>
+             {unreadNotifications.length > 9 ? '9+' : unreadNotifications?.length}
+           </span>
+          )
+        }
+         <Notifications style={{ color: colors.primary, fontSize: 24 }} />
```
NdibeRaymond, line 59: this too. move to notification style file

NdibeRaymond (changes requested):
test the functionality and it works @tuxology. Just a few comments and we can merge this @coderatomy . Nice work!

NdibeRaymond:
Something we should probably do in the future is to implement push notification. Right now users don't get notified unless they perform some kind of action that results in the frontend hitting the notification endpoint

tuxology:
@coderatomy Ok, I see that there was an update to the design 🤔 Its fine, we go with this design then. Note that the icon+notification bubble together is centered and has consistent padding along the sides so it looks balanced. In your implementation we see the notification bubble touching the outer circle edge. In addition, lets do changes requested by @NdibeRaymond as well Thanks @NdibeRaymond for testing. We'll merge this soon.

PR #997, opened 2023-11-12
(feat) Add login modal when searching projects for non-authenticated users

### LINKED ISSUE #802: Search is broken for non-logged in users

coderatomy:
## Summary
Adds login modal for non-authenticated users
Closes #802

## Changes
- Adds support for modalising the Login component
- Non-authenticated users can only search through the `featured projects`

### DIFF
--- a/zubhub_frontend/zubhub/src/views/search_results/SearchResults.jsx
+++ b/zubhub_frontend/zubhub/src/views/search_results/SearchResults.jsx
@@ -108,7 +108,6 @@ const buildCreatorProfiles = (
  * @todo - describe function's signature
  */
 function SearchResults(props) {
-  console.log(props, "PROPS")
   const classes = useStyles();
   const modalClasses = useModalStyles()
   const common_classes = useCommonStyles();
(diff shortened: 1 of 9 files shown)


tuxology:
Hey @coderatomy, thanks for redoing this 👍 I tried to test this and was unable to get the featured projects show up on test instance (behind the modal) when I searched as non-logged in user. Here is an example: 

image

The project does exist and is marked as featured project 👇 

image

zubhub_frontend/zubhub/src/views/search_results/SearchResults.jsx
```diff
@@ -282,6 +330,9 @@ const mapStateToProps = state => {
 
 const mapDispatchToProps = dispatch => {
   return {
+    getStaffPicks: args => {
```
tuxology, line 333: I see that we are only getting staff picks, but even when staff pics are enabled, we are not rendering. What could be the issue? Can you test locally and make changes if needed?

tuxology (changes requested):

Issue #1002, opened 2023-11-27
ZubHub Staging and Testing


NdibeRaymond:
## Requirements

- [x] Test a PR submitted by a contributor before merging and running on production

* Ability to test individual PRs where we can just completely run the whole ZubHub stack on a single PR/branch and get a URL to do manual testing
* Ability to see if migrations will go smoothly and any changes to backend will not affect API behavior

- [ ] Staging Instance
* Ability to have a publicly accessible (or via VPC for zubhub admins) staging instance on which we can do final testing for a “few days” before doing a final production release. 
* This staging instance can have multiple features/PRs already merged and staged together. Individual contributions will be “stacked” on this branch before release
* Testing/staging deployment should be available via Github Actions ideally

PR #1017, opened 2023-12-12
Create new github action to handle auto creation and destruction of test VMs

NdibeRaymond:
see #1002


NdibeRaymond:
hello @yokwejuste I deleted your suggestions on the previous patch (closed now) because that wasn't the focus of the PR. However, there are a number of actions that your suggestion will be beneficial to (including this one after we are certain it works as expected) so I think you should create an issue for `improving the readability of our github actions`. When you do, assign that me. I'd prefer to be the one making changes to those actions

yokwejuste:
> hello @yokwejuste I deleted your suggestions on the previous patch (closed now) because that wasn't the focus of the PR. However, there are a number of actions that your suggestion will be beneficial to (including this one after we are certain it works as expected) so I think you should create an issue for `improving the readability of our github actions`. When you do, assign that me. I'd prefer to be the one making changes to those actions

Definitely!

PR #1020, opened 2023-12-12
Improve more projects section

### LINKED ISSUE #1013: 'More Projects' section only shows the same user's projects
“More projects” section under project details only shows other projects by the same user instead of the intended behavior of showing similar projects from other creators. 

Screenshot 2023-12-11 at 13 46 09

This also results in showing an empty 'more projects' section when the user has no other projects, or showing the current project in this section.

Screenshot 2023-12-11 at 13 46 43

brrkrmn:
## Summary

'More Projects' section used to show the projects of the same user instead of showing similar projects from any user. With this PR, this section sends a request to the backend with the current project's id and retrieves 3 recommended projects depending on the project's categories. 

Closes #1013 

The changes in backend is adapted from @anthonycruzmacedo 's code in #458 

### Frontend
- api function is created to send a request to `projects/${id}/recommend/`
- error cases handled in projectActions.js with `getMoreProjects` function
- `projectDetails.jsx` uses `getMoreProjects`  to render the recommended projects

### Backend
- `'<uuid:pk>/recommend/'` url is used to recommend projects
- `ProjectRecommendAPIView` class returns 3 recommended projects 
- `recommend_projects` function handles the logic to determine the 3 projects by checking the projects with the same category/categories, and show the most liked projects if there's none.

### Before 
- shows the same user
Screenshot 2023-12-12 at 23 33 16
 - shows empty section when projects not found
Screenshot 2023-12-12 at 23 42 01
### After 
- shows other projects with the same categories from any user
Screenshot 2023-12-12 at 23 40 15
- shows most liked projects when no other project with the same category is found
Screenshot 2023-12-12 at 23 42 59

## Update
With the updated function, the projects are filtered based on their title, tags and categories. When doing title check, the project title is converted into an array of strings and each word is mapped. If any of the words are found in other titles, the project is appended. I will refer to these projects with 'similar title' in this PR.

**First loop is done with the similar titled projects, looking for:**
- Similar title, 1 matching category and 1 matching tag
- (if above filter's results are empty:) Similar title, 1 matching category or tag
- (if above filter's results are empty:) Similar title

**Second loop looks for projects with:**
- 1 matching category and 1 matching tag,
- (if above filter's results are empty:) 1 matching category or 1 matching tag

**Third loop looks for the most liked projects**


zubhub_backend/zubhub/projects/utils.py
```diff
@@ -463,3 +463,59 @@ def perform_project_search(user, query_string, search_criteria: Optional[Set[Pro
...
+                if len(projects) == 3:
+                  return projects
+
+    # looks for:
+    #  most likes
+    if len(projects) < 3:
+        for p in Project.objects.all().filter(publish__type=PublishingRule.PUBLIC).exclude(pk=project.pk).all().order_by('-likes_count'):
+            if not p in projects:
+                projects.append(p)
+                if len(projects) == 3:
```
yokwejuste: Some suggestions:
- Reduce Database Queries: Multiple calls to Project.objects.all() can be expensive. Instead, fetch the data once and filter it as needed.
- Improve Query Efficiency: Use select_related or prefetch_related for related objects (categories, tags) to reduce the number of database hits.
- Optimize Filters and Exclusions: Consolidate your filters and exclusions to reduce redundant iterations.
- Enhance the similarity check
- Ensure that the algorithm gracefully handles cases where there are fewer than 3 projects to recommend.
- Also about the code clarity.


I was thinking of something like this.
```python
def recommend_projects(project):
    Project = apps.get_model('projects.Project')
    PublishingRule = apps.get_model('projects.PublishingRule')

    title_words = set(project.title.split())
    categories = project.category.all()
    tags = project.tags.all()

    # Fetch projects excluding the current project and only public projects
    all_projects = Project.objects.filter(publish__type=PublishingRule.PUBLIC).exclude(pk=project.pk)

    # Similar Title, Same Category and Tags
    similar_title = Q(title__icontains=any(title_words))
    same_category = Q(category__in=categories)
    same_tags = Q(tags__in=tags)
    projects = all_projects.filter(similar_title & (same_category | same_tags)).distinct()

    # If not enough projects, add based on category and tags
    if projects.count() < 3:
        projects |= all_projects.filter(same_category & same_tags).distinct()

    # If still not enough, sort remaining by likes
    if projects.count() < 3:
        more_projects = all_projects.annotate(likes_count=Count('likes')).order_by('-likes_count')
        projects |= more_projects[:3 - projects.count()]

    return list(projects[:3])
```
This can make optimize it a little bit.

zubhub_backend/zubhub/projects/views.py
```diff
@@ -293,6 +293,24 @@ def get_object(self):
...
+    """
+
+    serializer_class = ProjectSerializer
+    permission_classes = [AllowAny]
+    throttle_classes = [CustomUserRateThrottle, SustainedRateThrottle]
+        
+
+    def get_queryset(self):
+        pk = self.kwargs.get("pk")
+        return recommend_projects(Project.objects.get(pk=pk))
```
yokwejuste: How do you handle a situation where you have a wrong id passed to the function? This will lead to a break, so we can foresee that and use the `get_or_404` function.

```python
        project = get_object_or_404(Project, pk=pk)

        try:
            return recommend_projects(project)
        except Exception as e:
            # Handle exceptions
            return Project.objects.none()
```

zubhub_frontend/zubhub/src/store/actions/projectActions.js
```diff
@@ -208,6 +208,33 @@ export const getProjects = args => {
...
+          };
+        } else {
+          res = Object.keys(res)
+            .map(key => res[key])
+            .join('\n');
+          throw new Error(res);
+        }
+      })
+      .catch(error => {
+        if (error.message.startsWith('Unexpected')) {
```
yokwejuste: The string check here can go wrong if translated, maybe focusing on the error message code can be better

yokwejuste (changes requested):
.

zubhub_frontend/zubhub/src/store/actions/projectActions.js
brrkrmn: Unexpected errors in catch blocks are handled like this in the rest of the project (around 77 times). I think for the sake of consistency this should stay like this here as well, but we can open an issue about it addressing all of them

yokwejuste: okay! Roger that.

yokwejuste (approved):

yokwejuste (approved):
LGTM 🚀

NdibeRaymond:
the code looks good. Still need to manually test the PR

PR #1032, opened 2023-12-14
fix: fixes the slash issue

### LINKED ISSUE #1031: fix slash issues on fetch

yokwejuste:
Closes #1031 

image


tuxology (changes requested):
Remove IDE config

PR #1050, opened 2023-12-21
Improve test coverage for Activities utilities

Chiemezuo:
## Summary

This PR adds tests for all `utils.py` functions for the `activities` app.
It is a step in the direction of improving code coverage across the entire project.

PS: I had to delete the `tests.py` file at the root of the `activities` app for the test folder to be recognized.


zubhub_backend/zubhub/activities/tests/test_utils.py
```diff
@@ -0,0 +1,223 @@
...
+
+class TestUtils(TestCase):
+  def setUp(self):
+    # sample image url from the internet
+    self.test_image_data = {
+      'file_url': 'https://picsum.photos/300',
+      'public_id': 'A sample image from the loren picsum website',
+    }
+    
+    # create image instance
```
NdibeRaymond, line 13: I don't think we should attempt to create the image instance ourselves. I think that is part of the things that `create_inspiring_artist` is expected to do if you take a closer look at the function. I think it should be  enough to pass `inspiring_artist_data` to `create_inspiring_artist` function (with the `image` field being set to `self.test_image_data`)

zubhub_backend/zubhub/activities/tests/test_utils.py
```diff
@@ -0,0 +1,223 @@
+from django.test import TestCase
+from activities.models import * #import all the models
+from activities.utils import * #import all the utilities
+
+class TestUtils(TestCase):
```
NdibeRaymond, line 5: this class should not be named `TestUtils` since that suggests that all the functions in the `utils.py` file are being tested under this `TestUtils` class. Although that is the approach you followed, it is wrong. What we should do instead is to have a different test class for each function in the utils.py file. for example `class TestCreateInspiringArtist, class TestUpdateImage, etc`. This is because for example for `update_image` we'd probably need to test more than one condition like when `image is truthy and image_data is truthy`, `image is truthy and image_data is falsy`, `image is falsy and image_data is truthy`, `image is falsy and image_data is falsy`

zubhub_backend/zubhub/activities/tests/test_utils.py
```diff
@@ -0,0 +1,223 @@
...
+    artist = create_inspiring_artist(inspiring_artist_data)
+    
+    self.assertIsInstance(artist, InspiringArtist)
+    self.assertEqual(artist.name, inspiring_artist_data['name'])
+    self.assertNotEqual(artist.name, 'Jane Doe')
+    self.assertEqual(artist.image, inspiring_artist_data['image'])
+    self.assertTrue(InspiringArtist.objects.filter(name='John Doe').exists())
+
+
+  def test_update_image(self):
```
NdibeRaymond, line 49: here you are testing only for one condition, when `image exists (truthy)` and `image_data is also provided and has necessary values (truthy)`. See my comment about class `TestUtils`.
Note that when I say that a variable is truthy, it means that evaluating that variable with an `if` conditional will result in `true`, and verse versa

zubhub_backend/zubhub/activities/tests/test_utils.py
```diff
@@ -0,0 +1,223 @@
...
+      {
+        'image': {
+          'file_url': 'https://picsum.photos/600',
+          'public_id': 'Activity image 2',
+        }
+      }
+    ]
+    
+    # get activity images count & images count before function
+    activity_image_count_before = ActivityImage.objects.filter(activity=self.test_activity).count()
```
NdibeRaymond, line 80: prev_activity_image_count

zubhub_backend/zubhub/activities/tests/test_utils.py
```diff
@@ -0,0 +1,223 @@
...
+    image_count_before = Image.objects.all().count()
```
NdibeRaymond, line 81: prev_image_count

zubhub_backend/zubhub/activities/tests/test_utils.py
```diff
@@ -0,0 +1,223 @@
...
+    # create activity images
+    create_activity_images(activity=self.test_activity, images=images)
+    # get activity images count & images count after function
+    activity_image_count_after = ActivityImage.objects.filter(activity=self.test_activity).count()
```
NdibeRaymond, line 85: activity_image_count

zubhub_backend/zubhub/activities/tests/test_utils.py
```diff
@@ -0,0 +1,223 @@
...
+    image_count_after = Image.objects.all().count()
```
NdibeRaymond, line 86: image_count

zubhub_backend/zubhub/activities/tests/test_utils.py
```diff
@@ -0,0 +1,223 @@
...
+      {
+        'title': 'Step 3',
+        'description': 'Description for step 3',
+        'step_order': 3,
+        'image': [{'file_url': 'https://picsum.photos/730','public_id': '3',}]
+      },
+    ]
+    
+    # get counts before calling function
+    activity_steps_count_before = ActivityMakingStep.objects.all().count()
```
NdibeRaymond, line 114: you can also use `old_acitivity_steps_count` and `new_activity_steps_count` instead of `activity_steps_count_before` and `activity_steps_count_after`. old and new conveys the information better

zubhub_backend/zubhub/activities/tests/test_utils.py
```diff
@@ -0,0 +1,223 @@
...
+    # call function to be tested
+    create_making_steps(self.test_activity, making_steps)
+    # get counts after function call
+    activity_steps_count_after = ActivityMakingStep.objects.all().count()
+    image_count_after = Image.objects.all().count()
+    self.assertEquals(activity_steps_count_after - activity_steps_count_before, len(making_steps))
+    # check that there are now more images than there were before
+    self.assertGreater(image_count_after, image_count_before)
+    
+  def test_create_inspiring_example(self):
```
NdibeRaymond, line 125: you might also want to test the image field of `Inspiring example 2`  InspiringExample model

zubhub_backend/zubhub/activities/tests/test_utils.py
```diff
@@ -0,0 +1,223 @@
...
+    create_inspiring_examples(self.test_activity, inspiring_examples=inspiring_examples_data)
+    # get number of examples and images after function call
+    examples_count_after = InspiringExample.objects.filter(activity=self.test_activity).count()
+    images_count_after = Image.objects.all().count()
+    # Assert new InspiringExample instances were created
+    self.assertGreater(examples_count_after, examples_count_before)
+    # Assert only one image was created
+    self.assertEqual(images_count_after - images_count_before, 1)
+
+  def test_update_activity_image(self):
```
NdibeRaymond, line 155: here you need to create some initial `ActivityImages` since we are testing an update operation.
We also need to mock out `create_activity_images` (this is to make sure that `update_activity_image` test can't be affected when change is made to `create_activity_images`. You can read more about mocking online).

Now after running `update_activity_image`, we check that the old `ActivityImages` that we manually created earlier no longer exists,
and check that the contents of the `images` array were passed to our mock.

zubhub_backend/zubhub/activities/tests/test_utils.py
```diff
@@ -0,0 +1,223 @@
...
+        }
+      }
+    ]
+
+    update_activity_images(self.test_activity, images_to_save=images)
+    # get count of activity images after function call
+    activity_images_count_after = ActivityImage.objects.filter(activity=self.test_activity).count()
+    self.assertEqual(activity_images_count_after, len(images))
+    
+  def test_update_making_steps(self):
```
NdibeRaymond, line 176: here you need to create some initial `ActivityMakingStep` manually since we are testing an update operation.
We also need to mock out `create_making_steps` (this is to make sure that `update_making_steps` test can't be affected when change is made to `create_making_steps`. You can read more about mocking online).

Now after running `update_making_steps`, we check that the old `ActivityMakingStep` that we manually created earlier no longer exists,
 and that the contents of  `making_steps` array were passed to our mock

zubhub_backend/zubhub/activities/tests/test_utils.py
```diff
@@ -0,0 +1,223 @@
...
+        'image': [{'file_url': 'https://picsum.photos/730','public_id': '3',}]
+      },
+    ]
+    
+    update_making_steps(self.test_activity, making_steps)
+    # get counts after function call
+    activity_steps_count_after = ActivityMakingStep.objects.filter(activity=self.test_activity).count()
+    self.assertEqual(activity_steps_count_after, len(making_steps))
+
+  def test_update_inspiring_examples(self):
```
NdibeRaymond, line 203: here you need to create some initial `InspiringExample` manually since we are testing an update operation.
We also need to mock out `create_inspiring_examples` (this is to make sure that `update_inspiring_examples` test can't be affected when change is made to `create_inspiring_examples`. You can read more about mocking online).

Now after running `update_inspiring_examples`, we check that the old `InspiringExample` that we manually created earlier no longer exists,
and that the contents of `inspiring_examples_data` array were passed to our mock

NdibeRaymond (changes requested):

zubhub_backend/zubhub/activities/tests/test_utils.py
Chiemezuo: I truly appreciate your feedback on this. I understand the idea behind what you've said. Earlier on, I also considered doing this, but a slight problem with this approach is that every class will need to have its own setup (to avoid dependence on another class). This would lead to a significant amount of repetition, and it would also be harder to maintain in the long run: for example, if something is changed about how activities or images are created in the `utils.py`, every test class that has an activity or image in its setup would have to be modified accordingly.

Still following your approach though, would it be better if I grouped into two classes: `class TestActivityOperations` and `TestImageOperations`? It would allow shared setups.
This would also prevent making classes out of minor `util` functions

zubhub_backend/zubhub/activities/tests/test_utils.py
Chiemezuo: I agree. This is a much more descriptive term. The change will reflect in my next PR

zubhub_backend/zubhub/activities/tests/test_utils.py
```diff
@@ -0,0 +1,223 @@
+from django.test import TestCase
+from activities.models import * #import all the models
+from activities.utils import * #import all the utilities
+
+class TestUtils(TestCase):
+  def setUp(self):
```
yokwejuste, line 6: function name in python are preferable when in snake casing and not Camel Casing or lower camel casing like in your case.

zubhub_backend/zubhub/activities/tests/test_utils.py
```diff
@@ -0,0 +1,223 @@
...
+        'image': [{'file_url': 'https://picsum.photos/720','public_id': '2',}]
+      },
+      {
+        'title': 'Step 3',
+        'description': 'Description for step 3',
+        'step_order': 3,
+        'image': [{'file_url': 'https://picsum.photos/730','public_id': '3',}]
+      },
+    ]
+
```
yokwejuste, line 112: When performing tests we need to have the images and other items created as instances of our models at runtime.

yokwejuste (changes requested):
Some changed needed

zubhub_backend/zubhub/activities/tests/test_utils.py
Chiemezuo: For the most part, yes. However, note [this](https://docs.djangoproject.com/en/5.0/topics/testing/overview/#writing-tests) example in the **official** Django documentation website.

PR #1082, opened 2024-01-07
feat: add tags to creator registration

### LINKED ISSUE #1081: new auth flow implementation from new design
From the new auth flow provided on the design we should implement some changes on the backend to match all this data.
image
image
image

yokwejuste:
## Summary

Description of PR here...
Closes #1081

Add tags in the request body to the creator's tags.

image

### DIFF
--- a/zubhub_backend/zubhub/creators/serializers.py
+++ b/zubhub_backend/zubhub/creators/serializers.py
@@ -199,8 +199,7 @@ class CustomRegisterSerializer(RegisterSerializer):
     bio = serializers.CharField(allow_blank=True, default="", max_length=255)
     subscribe = serializers.BooleanField(default=False)
     creator_tags = serializers.SlugRelatedField(slug_field='name',
-                                                queryset=CreatorTag.objects.all(),
-                                                many=True)
+                                                queryset=CreatorTag.objects.all())
 
     def validate_username(self, username):
         # Check if the value is a valid email
@@ -249,6 +248,14 @@ class CustomRegisterSerializer(RegisterSerializer):
             raise serializers.ValidationError(_("Location is required"))
         return location
 
+    def validate_creator_tags(self, creator_tags):
+        if (len(creator_tags) < 1):
+            raise serializers.ValidationError(_("Creator tags are required"))
+        for tag in creator_tags:
+            if not CreatorTag.objects.filter(name=tag).exists():
+                raise serializers.ValidationError(_("Invalid creator tag: {}".format(tag)))
+        return creator_tags
+
     def get_cleaned_data(self):
         data_dict = super().get_cleaned_data()
         data_dict['phone'] = self.validated_data.get('phone', '')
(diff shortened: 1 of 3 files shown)


zubhub_backend/zubhub/creators/serializers.py
```diff
@@ -259,6 +261,7 @@ def get_cleaned_data(self):
 
     def save(self, request):
         creator = super().save(request)
+        creator.tags.set(self.validated_data.get('creator_tags', []))
```
NdibeRaymond: this shouldn't be done here @yokwejuste. You should take a second look at my little pointers I commented on the issue itself.

NdibeRaymond: * First the tag here should be a string and not an array of strings (I am not sure we expect more than one tag on the frontend during registration do we?)
* getting tag should be done in `get_cleaned_data` function above and set in the adapter.
* Also remember we need custom validation for the submitted tag. We can't trust the tags submitted through the api since anything can be submitted. We need to validate it both on the frontend and on the backend.
One way to validate it is to check that the submitted tag string matches the name of one of the tags we have on the backend.

zubhub_backend/zubhub/creators/serializers.py
```diff
@@ -199,6 +198,9 @@ class CustomRegisterSerializer(RegisterSerializer):
                                             queryset=Location.objects.all())
     bio = serializers.CharField(allow_blank=True, default="", max_length=255)
     subscribe = serializers.BooleanField(default=False)
+    creator_tags = serializers.SlugRelatedField(slug_field='name',
+                                                queryset=CreatorTag.objects.all(),
+                                                many=True)
```
NdibeRaymond: we'd need to adjust this because we expect just a single tag string (again if you think there is a reason we need to expect an array please I'd like to hear)

zubhub_backend/zubhub/creators/views.py
```diff
@@ -188,6 +188,7 @@ class RegisterCreatorAPIView(RegisterView):
             "location": "string",\n
             "bio": "",\n
             "subscribe": false\n
+            "creator_tags": ["tag1", "tag2", ...]\n
```
NdibeRaymond: yeaa this is good. If we attend to the other reviews above then we'd need to modify this too

NdibeRaymond (changes requested):

zubhub_backend/zubhub/creators/serializers.py
yokwejuste: moved to a string, the Array was in case we wanted to create a user with several roles, but it is okay we can do that from the Django admin panel

zubhub_backend/zubhub/creators/serializers.py
yokwejuste: copied, all implemented

NdibeRaymond (approved):

PR #1087, opened 2024-01-19
(feat): Improve search for non-logged-in users

### LINKED ISSUE #802: Search is broken for non-logged in users

brrkrmn:
## Summary

This PR addresses the intended flow explained [here](https://github.com/unstructuredstudio/zubhub/issues/802#issuecomment-1783577585). Now, logged-out users can search for projects and creators 

Closes #802

## Changes
Although this PR is to make changes for non-logged-in users, I kept some of the changes that also affect logged-in user's experience but can remove them upon your comments. 

**For non-logged-in users:**
- Search dropdown is disabled
- If the query matches with multiple results, they only see 3 of them at most
- they see a login form when they make search with or without results
- If there are no results, 3 staff pick projects are shown instead (only for project search)

**For logged-in users:**
- They also see the staff pick projects when their project search yields no results
- When they make a creator search with no results, they see the same page with  non-logged-in users except for the login form

## To Do

- [ ] Make the results' paths directly accessible by non-loggers but limit the other projects/creators

## Screenshots

Please see how results look for different combinations [here](https://www.figma.com/file/U27BHBSAA8d91ePGYLVQrM/Search?type=design&node-id=0%3A1&mode=design&t=LkPxArBtJTdci83m-1)

Here is a quick demonstration of how search looks like for an unauthenticated user:

https://github.com/unstructuredstudio/zubhub/assets/92817363/b752b529-ce30-43d1-ad26-f114abca8c72

## Further Improvement

There are some things to be improved for the overall search feature, which can be a part of another PR. Some of them are:
- Currently, tag search is not possible
- Styles should be made more unified since there are a lot of inconsistencies
- Searching for projects or creators is not possible unless the user types a whole word (e.g. searching for '' does not work)


brrkrmn:
### Update: results not overflowing 1 row for non-logged-in users

zubhub_frontend/zubhub/public/locales/en/translation.json
NdibeRaymond, line 1: for the Sanskrit translation, I have a pro-tip:
Since we already have the english sentence,  we can copy it, search "translate english to sanskrit" on google, paste the english sentence in the translation form and get the sanskrit from the other

NdibeRaymond:
code looks ok but still need to test the UI with user data

NdibeRaymond:
had to refactor a lot of things here (some of which are not directly related to this patch but still hurts my eyes). It's good to go now

PR #1091, opened 2024-01-29
Design activity card

### LINKED ISSUE #1090: (activity): Redesign activity card
The [new design](https://www.figma.com/file/nyUduEpFmVZ05xA8rt28Mt/ZubHub-Feature-Re-Designs?type=design&node-id=835-9115&mode=design&t=aUFMUZG0tfebun9Q-0) for activity component needs to be implemented.
Published and unpublished activity card design: https://www.figma.com/file/nyUduEpFmVZ05xA8rt28Mt/ZubHub-Feature-Re-Designs?type=design&node-id=304-424&mode=design&t=FsRULrVG2GYIKguh-0

brrkrmn:
## Summary

The activity component is redesigned according to the [new proposed design](https://www.figma.com/file/nyUduEpFmVZ05xA8rt28Mt/ZubHub-Feature-Re-Designs?type=design&node-id=304-424&mode=design&t=fGqjo466CWeAlZmp-0) .

Closes #1090 

## Changes

- If there are multiple creators of an activity, their avatars get stacked in the card. 3 creators get stacked at most, after which the number of remaining users are displayed in the last avatar. On desktop, overlapped avatars get separated on hover (see video in the section).
- Category section is separated into its own component, using MUI Chip component. Each category has a unique background color that is accessed from the `colors.js` file.
- The introduction of activity is displayed under the title.
- User's tag is placed under the username.

### Desktop
desktop

### Multiple Creators Demonstration

https://github.com/unstructuredstudio/zubhub/assets/92817363/8a2b6e4d-8f1e-4f57-8764-a03359b73779


zubhub_frontend/zubhub/src/components/activity/activity.jsx
```diff
@@ -63,128 +47,137 @@ function Activity(props) {
...
+                            className={classes.creatorUsername}
+                          >
+                            {creator.username}
+                          </Typography>
+                          <Typography
+                            color="textPrimary"
+                            className={classes.creatorTag} component="p"
+                          >
+                            {creator.tags[0]}
+                          </Typography>
```
chaitanya-cloudknox: loop through tags instead, incase if we support multiple tags it just works.

zubhub_frontend/zubhub/src/components/activity/activity.jsx
```diff
@@ -63,128 +47,137 @@ function Activity(props) {
...
-                </Grid>
-              </Grid>
+                ))
+                : (
+                  <Box className={classes.creatorBox}>
+                    <Box className={clsx(activity.creators?.length === 2 ? classes.twoCreatorBox : classes.multipleCreatorBox)}>
+                      {activity.creators.slice(0, 3).map((creator, index) => (
+                        <Tooltip
+                          key={creator.id}
+                          title={index === 2 && activity.creators.length > 3 ? `${activity.creators.length - 2} more` : creator.username}
```
chaitanya-cloudknox: title needs to be translated

zubhub_frontend/zubhub/src/components/activity/activity.jsx
```diff
@@ -63,128 +47,137 @@ function Activity(props) {
...
+                          placement="bottom"
+                          arrow
+                          className={classes.creatorUsernameTooltip}
+                          style={{ zIndex: index }}
```
chaitanya-cloudknox: Why do we need different z-indexes? What's issue with same index?

zubhub_frontend/zubhub/src/components/categories/Categories.jsx
```diff
@@ -0,0 +1,24 @@
+import { makeStyles } from "@material-ui/core";
+import { styles } from "./categories.styles";
+import Chip from '@material-ui/core/Chip';
+import { categoryColors } from "../../assets/js/colors"
```
chaitanya-cloudknox: Its good idea to add category colors in the theme and consume using useTheme

NdibeRaymond:
@brrkrmn what happens if we have much more creators (maybe 10 or more)? can you attempt this and make a video so we see how the hover effect works in that case?
Wondering if it will overflow or wrap. This should also be addressed in the design, was it?

zubhub_frontend/zubhub/src/assets/js/colors.js
NdibeRaymond, line 1: in the future we might want to move all our colors into this file (you can create a task for this if you want)

NdibeRaymond:
needs rebase

brrkrmn:
> @brrkrmn what happens if we have much more creators (maybe 10 or more)? can you attempt this and make a video so we see how the hover effect works in that case? Wondering if it will overflow or wrap. This should also be addressed in the design, was it?

It'll look like what we have for +3 creators, showing 3 circles at most and indicating the number of the remaining creators. Like this: 

https://github.com/unstructuredstudio/zubhub/assets/92817363/93c61a23-7bce-4afd-b711-3ac3ee6de1ff

This will only be applicable when the 'team' feature is made available. I'm not sure how teams are going to look exactly but if they have their own avatars, we can use the name and avatar of the team instead of showing all the creators in the activity card. And then we can show all creators in detail in the team's page or activity details page. In that case, we should have a larger section for the creators to be able to show this many.

NdibeRaymond:
@tuxology this has been rebased so we can focus on testing before merging to staging

PR #1096, opened 2024-02-03
Redesign Activity Details Page

brrkrmn:
## Summary

Redesigned activity details page according to the [new design](https://www.figma.com/file/nyUduEpFmVZ05xA8rt28Mt/ZubHub-Feature-Re-Designs?type=design&node-id=304-424&mode=design&t=LS8mVmTD3LVCmyfD-0). 

Depends on #1091 

Closes #1092 

## Changes

- Activity actions menu is shown if the user is the creator of the activity or a staff member.
- Activity actions menu shows 'publish'/ 'unpublish' button if the user is a staff member.

## To Do

- [ ] Use `creators` component to show activity creators
- [x] Make 'publish' and 'unpublish' buttons functional

PR #1104, opened 2024-02-15
(activity): Activity creation changes

### LINKED ISSUE #1097: (activity): Activity Creation
This issue presents the goals for the Activity Feature, focusing on the changes in **Activity Creation** process.
Changes needed for limiting creator’s access to this form is included in the issue #1099  

## Checklist:

- [ ] The [new design for the activity creation form](https://www.figma.com/file/nyUduEpFmVZ05xA8rt28Mt/ZubHub-Feature-Re-Designs?type=design&node-id=971-2857&mode=design&t=YLuFiVMxHBQIATO9-0) should be implemented.
- [x] Users with 'Educator' role should be able to fill Activity Creation Form without any issues. If there is a field missing the UX should accurately inform them what is to be done.
- [x] In activity creation form, remove the 'Preview' and 'Saved to Draft' indication from the secondary nav. That is misleading since we don't save activity in drafts.
- [x] Once an Educator creates an activity, it should be saved as unpublished by default.

## Additional Issues/Concerns

- [ ] In the Activity Creation Form, the text fields lose focus as the user types in and prevents them to type more. The user must click on the field to continue typing.

https://github.com/unstructuredstudio/zubhub/assets/92817363/0fab6225-45ab-47af-b5d0-e6e3a6ea62d3

- [x] In the modal that pops up after the user creates an activity, we should let them know that the activity is not yet published and it will be accessible by registered and unregistered people once it is published.

Screenshot 2024-02-11 at 1 46 10 PM

- [x] The submit button in the Activity Creation Form's last step should be labeled as 'Create' instead of 'Publish' because the Publishing is a different step made by staff members, so it can be misleading.

Screenshot 2024-02-11 at 1 47 56 PM

brrkrmn:
## Summary

This PR addresses the changes needed for the activity creation process. The prioritized tasks were functional needs so the new design is not yet implemented.

Closes #1097 

## Changes

1. Created activities are unpublished by default
2. Removed the 'preview' and 'draft' banner from activity creation form. The 'go back' arrow is also removed since it was not functional.
3. The label of the submit button in the last step is changed from 'publish' to 'create'
4. The modal that appears after the user creates an activity is edited:
- Removed social buttons because the user should not be able to share the activity until it is published
- Changed the texts to explain that the activity is created and in waitlist to be published.  

## Screenshots

Screenshot 2024-02-15 at 11 45 34 PM
Screenshot 2024-02-16 at 12 01 07 AM
Screenshot 2024-02-16 at 12 01 20 AM

### DIFF
--- a/zubhub_frontend/zubhub/src/views/activity_details/ActivityDetailsV2.jsx
+++ b/zubhub_frontend/zubhub/src/views/activity_details/ActivityDetailsV2.jsx
@@ -23,7 +23,6 @@ import ReactQuill from 'react-quill';
 import { useSelector } from 'react-redux';
 import { useReactToPrint } from 'react-to-print';
 import Html2Pdf from 'html2pdf.js';
-
 import ZubHubAPI from '../../api';
 import { colors } from '../../assets/js/colors';
 import { ClapBorderIcon } from '../../assets/js/icons/ClapIcon';
@@ -283,24 +282,21 @@ export default function ActivityDetailsV2(props) {
       </div>
 
       <Modal.WithIcon icon={<FiShare size={30} />} maxWidth="xs" open={open} onClose={toggleDialog}>
-        <div style={{ display: 'flex', justifyContent: 'end' }}>
+        <div className={classes.dialogCloseButton}>
           <IconButton onClick={toggleDialog}>
             <CloseOutlined />
           </IconButton>
         </div>
 
         <DialogTitle>
-          <Typography align="center" className={clsx(commonClasses.title2, classes.dialogTitle)}>
-            Congratulations your Activity has been successfully created!
+          <Typography align="center" className={classes.dialogTitle}>
+            {t('activityDetails.activity.create.modal.primary')}
           </Typography>
         </DialogTitle>
         <DialogContent>
-          <Typography align="center">
-            Share your activity with the world. Post it on the following platforms:
+          <Typography align="center" className={classes.dialogText}>
+            {t('activityDetails.activity.create.modal.secondary')}
           </Typography>
-          <div className={clsx(commonClasses.displayFlex, commonClasses.justifyCenter)} style={{ margin: '20px 0' }}>
-            <SocialButtons containerStyle={{ gap: 50 }} withColor link facebook whatsapp />
-          </div>
         </DialogContent>
       </Modal.WithIcon>
     </div>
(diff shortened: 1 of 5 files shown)


NdibeRaymond:
still needs rebasing right @brrkrmn ?

zubhub_frontend/zubhub/src/views/activity_details/ActivityDetailsV2.jsx
```diff
@@ -282,24 +282,21 @@ export default function ActivityDetailsV2(props) {
...
+        <div className={classes.dialogCloseButton}>
           <IconButton onClick={toggleDialog}>
             <CloseOutlined />
           </IconButton>
         </div>
 
         <DialogTitle>
-          <Typography align="center" className={clsx(commonClasses.title2, classes.dialogTitle)}>
-            Congratulations your Activity has been successfully created!
+          <Typography align="center" className={classes.dialogTitle}>
```
NdibeRaymond, line 292: thanks for doing this @brrkrmn ! seeing untranslated strings stresses me out a lotttttt

zubhub_frontend/zubhub/src/views/create_activity/script.js
```diff
@@ -673,7 +673,7 @@ export const submitForm = async ({ step1Values, step2Values, props, state, handl
     const formData = {
         ...formDataStep1,
         ...step === 2 ? step2Values : {},
-        publish: step === 2 ? true : false
+        publish: false
```
NdibeRaymond: I wonder why we are not making it possible for a user to publish an activity immediately it is created?

NdibeRaymond (commented):
few comments

zubhub_frontend/zubhub/src/views/activity_details/ActivityDetailsV2.jsx
brrkrmn: oh I know @NdibeRaymond , this part was edited in the [activity details page PR](https://github.com/unstructuredstudio/zubhub/pull/1096/files#diff-bacf14af98db90a1be92abb4487123821b095374989badbdd916e961b0ac4a4d) which uses all text from translation, so I didn't want to redo the work here 😅

zubhub_frontend/zubhub/src/views/create_activity/script.js
brrkrmn: after discussing this with @tuxology we've come to agree that it should be the staff members who publish activities after making any necessary changes. That's why we now added a whole new tab for unpublished activities for staff members to see. I assume it's because we want activities to be presented in a more monitored and controlled way than projects.

NdibeRaymond:
@tuxology this has been rebased so you can focus on testing before merging to staging.

PR #1109, opened 2024-02-20
Revert "feat: add tags to creator registration"

NdibeRaymond:
Reverts unstructuredstudio/zubhub#1082

there is an error in this PR that makes it impossible to register a new user. @yokwejuste you should look at this again and try to verify that it works before creating a PR again. I assumed you already did that so didn't test properly before merging. Possibly try to get the tests in `zubhub_backend/zubhub/creators/tests/test_views.py` to pass.

### DIFF
--- a/zubhub_backend/zubhub/creators/serializers.py
+++ b/zubhub_backend/zubhub/creators/serializers.py
@@ -8,7 +8,8 @@ from django.contrib.auth import get_user_model
 from django.utils.translation import gettext_lazy as _
 import csv
 from .admin import badges
-from .models import CreatorGroup, Creator, CreatorGroupMembership, Location, PhoneNumber, CreatorTag
+from .models import CreatorGroup, Creator, CreatorGroupMembership
+from .models import Location, PhoneNumber
 from allauth.account.models import EmailAddress
 from rest_auth.registration.serializers import RegisterSerializer
 from rest_auth.serializers import PasswordResetSerializer
@@ -198,8 +199,6 @@ class CustomRegisterSerializer(RegisterSerializer):
                                             queryset=Location.objects.all())
     bio = serializers.CharField(allow_blank=True, default="", max_length=255)
     subscribe = serializers.BooleanField(default=False)
-    creator_tags = serializers.SlugRelatedField(slug_field='name',
-                                                queryset=CreatorTag.objects.all())
 
     def validate_username(self, username):
         # Check if the value is a valid email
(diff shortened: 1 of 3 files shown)


yokwejuste:
Please this should work with @coderatomy new UI the signup

NdibeRaymond:
> Please this should work with @coderatomy new UI the signup

It does not work. I just tested it now

zubhub_backend/zubhub/creators/views.py
```diff
@@ -186,7 +188,6 @@ class RegisterCreatorAPIView(RegisterView):
             "location": "string",\n
             "bio": "",\n
             "subscribe": false\n
-            "creator_tags": "string",\n
```
NdibeRaymond, line 189: here you are specifying that `creator_tags` is a string, but you were treating it like an array everywhere else

zubhub_backend/zubhub/creators/serializers.py
```diff
@@ -248,22 +247,13 @@ def validate_location(self, location):
...
-        return creator_tags
-
     def get_cleaned_data(self):
         data_dict = super().get_cleaned_data()
         data_dict['phone'] = self.validated_data.get('phone', '')
         data_dict['dateOfBirth'] = self.validated_data.get('dateOfBirth', '')
         data_dict['location'] = self.validated_data.get('location', '')
         data_dict['bio'] = self.validated_data.get('bio', '')
         data_dict['subscribe'] = self.validated_data.get('subscribe', '')
-        data_dict['creator_tags'] = self.validated_data.get('creator_tags', [])
```
NdibeRaymond, line 266: here you treat `creator_tags` as an array

zubhub_backend/zubhub/creators/serializers.py
```diff
@@ -198,8 +199,6 @@ class CustomRegisterSerializer(RegisterSerializer):
                                             queryset=Location.objects.all())
     bio = serializers.CharField(allow_blank=True, default="", max_length=255)
     subscribe = serializers.BooleanField(default=False)
-    creator_tags = serializers.SlugRelatedField(slug_field='name',
```
NdibeRaymond, line 201: the syntax here is basically saying `Creator.objects.all().filter(name=<value of creator_tags coming from the frontend>)[0], or something like that`. But the value of the creator_tags coming from the frontend given the way we are trying to extract it and the default we are providing, is an array, so won't work

NdibeRaymond:
from the reviews above, you can see that you are mixing the `creator_tags` field. Treating it as an array in one place and treating it as string in others

PR #1120, opened 2024-03-01
use route hooks from the pagewrapper

### LINKED ISSUE #1121: Projects details view is not routed properly
**Describe the bug**
When we try to view any project (eg. https://zubhub.unstructured.studio/projects/0848243c-5381-45f3-bacb-e60c58fdda73) the request is redirected to https://zubhub.unstructured.studio/projects/undefined

**To Reproduce**
Go to Zubhub, list projects, then click on any project to view its detail. Address bar shows full route, https://zubhub.unstructured.studio/projects/0848243c-5381-45f3-bacb-e60c58fdda73 but the request is actually sent to https://zubhub.unstructured.studio/projects/undefined which returns 404

**Expected behavior**
We see project properly

**Screenshots**
image

coderatomy:
fixes https://github.com/unstructuredstudio/zubhub/issues/1121
fixes validation errors


coderatomy:
more work has to be done since `params` is extensively used https://github.com/search?q=repo%3Aunstructuredstudio%2Fzubhub%20props.params&type=code

zubhub_frontend/zubhub/src/views/create_project/script.js
```diff
@@ -693,21 +693,8 @@ export const validationSchema = Yup.object().shape({
...
-                value.split(',').forEach(material => {
-                    if (material) {
-                        is_empty = false;
-                    }
-                });
-
-            return !is_empty;
-        }),
-    category: Yup.string(),
+    materials_used: Yup.array().required('required'),
```
NdibeRaymond: No idea why we keep removing  validations instead of getting them to work properly

NdibeRaymond: If we are going to work on this then we should fix it completely

NdibeRaymond (changes requested):

NdibeRaymond:
> fixes #1121 fixes validation errors

* fixed  eslint  errors
*  improvements to App.js  and PageWrapper.jsx
* removed materials_used and category field validation issues. Those require their own issue and PR

PR #1122, opened 2024-03-02
Feat/new auth flow fix

yokwejuste:
## Summary
This PR is an implementation of the new auth flow.

Copy of #1081


NdibeRaymond:
You should also rebase, and fix the failing pre-commit changes. I posted the link to a video I made on zulip, about how to go about that

NdibeRaymond:
part of that failing pre-commit test is the little test we have on the backend for user creation. You need to update the test to make sure it passes. Right now it's failing, which is a pointer that the PR is failing. This looks to be the same PR that was reverted right?

PR #1126, opened 2024-03-20
fix: fixes errors

yokwejuste:
## Summary
Fixes the material used validation error during creation


tuxology (approved):
I am merging this one as an exception (even with failing pre-commit checks) since it is affecting the flow of new project creation @NdibeRaymond

